PR #1, opened 2026-08-03
Correct ADR-018's rationale in place; record when to version vs supersede

dmarx:
You were right about ADR-002, and correcting it turned out to need a rule that didn't exist.

## The mis-citation

[ADR-018](https://github.com/dmarx/luria/blob/claude/badge-rationale-and-pr-workflow/docs/decisions/ADR-018.md) rejected the endpoint-badge alternative by invoking ADR-002's per-merge bot commit. That over-applies it. ADR-002's hazard is specific:

- `CHANGELOG.md` is **appended to at a marker**, so a bot commit forces in-flight branches to rebase into a conflict *exactly where their own content goes*
- its entries carry **assigned numbers**, so a careless resolution silently drops somebody's entry

A derived badge file has neither. It's a wholesale rewrite of a computed value; a conflict in it is resolved by regenerating. Nothing is lost, nothing is numbered, nothing is appended.

Worth naming the failure mode, since it's the one this package exists to prevent in documentation and I committed it in reasoning: **a decision cited as a principle**. ADR-002 records a choice made against one file's shape. Quoting it as "bot commits are bad" strips the conditions that made it true — and a rule with its conditions stripped is what gets misapplied to the next problem. The record has a separate layer for standing values, and the fact that a claim didn't earn a place there is information.

## The decision still stands, for the reason I didn't write down

A baked-in URL is correct **per commit**. Open a PR that adds a `Proposed` decision and that branch's README already says `needs decision: 1`, in the diff, where a reviewer is looking. An endpoint badge always reports the default branch — so the one moment the number is most worth seeing, while deciding whether to merge the thing that moves it, is the moment it can't show it. Needing no CI and no second file is the smaller half of the argument, and I'd led with it.

That property is also what makes the badges worth having under the new PR workflow, which is a nice accident rather than foresight.

## Correcting it needed ADR-019

Read literally, "never rewrite a decision's body" leaves only supersession — retiring a decision still in force, repointing its citations at a near-identical replacement, and telling every future reader the *choice* changed when only an argument did. A few of those and `Superseded` stops meaning anything, which defeats the closed status vocabulary (ADR-003).

So [ADR-019](https://github.com/dmarx/luria/blob/claude/badge-rationale-and-pr-workflow/docs/decisions/ADR-019.md) splits the cases:

| what is wrong | remedy |
|---|---|
| the **choice** — we'd decide differently now | supersede |
| a **reason** — the choice stands, an argument for it doesn't | correct in place, bump `version`, say what changed in `history:` |

The test for the ambiguous case: *would a reader who acted on the old version have done something different?* If yes, supersede. If they'd have done the same thing for a worse reason, correct in place.

What makes it safe rather than a loophole: "never rewrite a body" is objecting to **silent** revision. A version bump plus a history note saying what the previous version got wrong is the opposite of silent. ADR-018 is `v2` accordingly — the first thing the version field has had to display for a decision, and it arrived a day after ADR-016 added it on a speculative argument.

## Also in this PR

- **Contributions go through a pull request from here on**, recorded in `CLAUDE.md` alongside the correction rule.
- **A cascade worth keeping**: creating ADR-019 made `ADR-019` a real code — and it had been serving as a *fixture* number in `doc_refs.py` and `test_doc_refs.py` under `unresolved-ok` annotations. The inverted check fired immediately (`annotation names ADR-019, which does resolve here`), and because a malformed annotation excuses nothing, the *other* code in the same annotation fell out of cover and got reported too. Two levels of consequence from one new file, both automatic. Fixture codes now live at 900+.

## Verification

152 tests pass · `luria lint` clean · `luria index --check` current · badges unchanged at 0/0.

---
_Generated by [Claude Code](https://claude.ai/code/session_014H7APrKgWA23SVZMNF8H86)_


dmarx:
Pushed a second commit to this branch: the docs now say plainly that the record is revisable, and point at where it already was.

**The problem ADR-019 alone didn't fix.** It settled *what* to do about a wrong reason, but "never rewrite a decision's body" was stated **absolutely in six places** — ADR-001's traffic rules, the doctrine page, both index stubs, both decision templates, `CLAUDE.md`. A rule stated absolutely in six places and refined in one gets applied absolutely, because nobody reads all seven. Every site now carries the same sentence rather than hedging in its own words: *the objection is to silent revision, not to editing.*

**`docs/project-memory.md` gains section 2**, with the four shapes and a live example of each from this repository — because abstract permission changes nobody's behaviour, and a list of things that already happened does:

| what happened | remedy | already here |
|---|---|---|
| the **choice** changed | supersede | ADR-010 → ADR-011; ADR-015 → ADR-016 (within hours) |
| a **reason** was wrong, choice stands | correct in place, `version` + `history:` | ADR-018 `v2` |
| a **value** widened or was reworded | same | DP-2 `v2`, DP-3 `v2` |
| a **consequence** stopped being true | a later document records the new state | ADR-016's consequence, falsified by ADR-017 |

That fourth shape had no name before writing this, and it's genuinely distinct — a consequence is an *observation*, and observations expire without the decision under them being wrong.

**ADR-001 is `v2`.** Its clause now says supersede when the **choice** changes, with `history:` recording that version 1 read as "documents are frozen". This is the mechanism proving itself on the document that stated the problem: the over-broad wording is fixed by exactly the procedure the fix introduces, and both versions are visible. Superseding a four-layer taxonomy over one clause of one bullet would have been absurd.

⚠️ **Two things to push back on if I've read you wrong:**
- Versioning **ADR-001** is me editing the foundational decision. Defensible under ADR-019, and still the sort of thing you should get to veto.
- I softened the *template* and *scaffold* copy too, so this changes what adopting projects are told, not just what this repo does.

**Also:** naming ADR-010 and ADR-015 as the supersession examples produced two unacknowledged retired-citation warnings — exactly what `inactive-ok` is for — and the `-block` scope covered the heading rather than the bullet list under it, so it had to widen to `-file`. Fixing that took the `cited but retired` badge from amber back to green in the same `luria index` run: paragraph written → number on the front page moved, one command.

152 tests · lint clean · `index --check` current · fresh scaffold clean.

---

PR #30, opened 2026-08-07
feat: improve documentation, speak some in my own voice

dmarx:


dmarx:
There's certainly more to do, but this is more than enough to merge. Maybe I'll add a short devlog entry first cause why not.

PR #32, opened 2026-08-07
Trim the CLI to the workflows: six contributor commands, two for CI (ADR-030)

dmarx:
The CLI had eleven commands because the dispatcher mapped one command per module — the package layout projected onto the interface, not anyone's workflow. Three of the names claimed workflows nobody had, and the codebase had already said so about itself: `reports.py`'s docstring calls `ref-status`/`pending` commands "which nobody runs", and bare `luria badges` printed a stderr warning that `luria index` is what you meant (the ADR-029 incident).

## What changed

- **`luria --help` is now tiered**: six contributor commands (`lint`, `link`, `index`, `journal`, `remotes`, `init`) and two labelled "run by CI, rarely by hand" (`reports`, `collect`).
- **`badges`, `ref-status`, and `pending` are retired.** A retired name exits 2 naming its successor instead of "unknown command" (DP-1) — e.g. `luria badges` answers "`luria index` writes the README's badges and `luria lint` checks them".
- **The modules keep their `main()`s**: `python -m luria.ref_status --all` remains the interactive escape hatch, and the vendor-one-file property is unchanged. Retiring a command deleted dict entries, not code.
- Every string that recommended a retired command was repointed — the lint's warning footers, `remotes`' report output, and the module docstrings now cite `luria reports`.
- Makefile drops the `ref-status`/`pending` targets; README, `CLAUDE.md`, and `template/CLAUDE.md` describe the same six-plus-two surface.
- New `tests/test_cli.py` covers the tiered help, the successor-naming refusals, and that no retired name is still registered.

## Record

- **ADR-030** documents the decision, including the rejected alternatives (keeping all eleven; folding `link` into `lint --fix`; deleting the modules' entry points; silent removal).
- **ADR-007** is corrected in place to v2 with a history note — its warn-don't-fail choice stands; only the delivery surface (the two standalone commands) moved to ADR-030.
- Changelog fragment and a devlog entry recording the dead ends, including the trap that retired command names live in *strings*, not just the dispatch table.

## Compatibility

An adopter with `luria badges` or `luria ref-status` in a script gets exit code 2 with the successor named on stderr — a break, but a self-explaining one. The ADR-029 composite actions (`actions/generate`, `actions/lint`) never used the retired names, so scaffolded workflows are unaffected.

`make ci`: 270 tests passing, docs lint clean. The badge counts in this PR's README reflect the branch's own record per ADR-018.

---
_Generated by [Claude Code](https://claude.ai/code/session_018AHHTjpaGcX2eQfE9kgvkd)_


luria/cli.py
```diff
@@ -2,19 +2,24 @@
...
     luria init          scaffold the record into a project that has none
 
+Two more exist for CI, which is their only regular caller:
+
+    luria reports       write the status reports as markdown, for an artifact
+    luria collect       assemble fragment directories into their views
+
+The surface used to be wider — one command per module, eleven in all — which
+mirrored the package layout rather than anyone's workflow (ADR-030). A retired
+command answers with where its job went, not "unknown command".
```
dmarx: This makes sense as a general rule of thumb, but with one caveat: deprecations should communicate a migration timeline. 

I've noticed that LLM development style tends to be biased towards preserving legacy functionality. I think it's great to be risk averse, but this can result in a lot of unnecessary complexity. This PR is a great example: this library is only a few days old and only has one customer. The whole point of this PR was clean up an API that had experienced feature creep. Who's workflow are we protecting by maintaining the existence of these deprecated commands?

Preserving legacy behavior makes sense if there's a real risk that a deprecation could interfere with someone's workflow, but otherwise it's ridiculous to preserve an unnecessarily wide API like this. It makes the API harder to use by rendering the commands we want to direct the user towards harder to find.

It's great to use an affordance to guide an undesirable behavior that's likely to occur towards a more desirable behavior. It's stupid to create an affordance in service of a behavior that wouldn't even exist if the affordance was just removed. Good: putting ashtrays in airplane bathrooms because we can't control if some asshole might try to smoke on the plane. Bad: passing out toy guns to passengers that when triggered, deploys a flag with the message "guns aren't permitted on this flight".

luria/cli.py
```diff
@@ -24,22 +29,39 @@
...
+# workflow needs them — `collect` even mildly misfires locally, consuming
+# fragments a reviewer was meant to see on the branch.
+CI_COMMANDS = {
+    "reports": ("luria.reports", "write the status reports as markdown"),
+    "collect": ("luria.collect", "assemble fragments into their views"),
+}
+
+# Each of these was subsumed before it was retired (ADR-030); the refusal
+# names the successor rather than pleading ignorance (DP-1).
+RETIRED = {
```
dmarx: let's just *actually* deprecate this stuff. see rant above.

luria/cli.py
```diff
@@ -24,22 +29,39 @@
 COMMANDS = {
```
dmarx, line 27: this whole thing could be significantly simplified by using the Fire library.

luria/cli.py
```diff
@@ -2,19 +2,24 @@
...
+
 Subcommands delegate to modules that each keep their own `main()`, so any of
-them still runs standalone — useful when a project vendors one file instead of
-installing the package.
+them still runs standalone (`python -m luria.ref_status`) — useful when a
+project vendors one file instead of installing the package, and how the
+retired reports stay reachable in full detail.
 """
 
 from __future__ import annotations
```
dmarx: this is unnecessary.

dmarx: uh correction... actually, it's *wrong*. use `annotationlib` instead. see: https://peps.python.org/pep-0749/

record/decisions.d/ADR-007.md
```diff
@@ -1,13 +1,26 @@
...
+  note: >-
+    ADR-030 retired the standalone commands nobody ran; the reports now reach
+    readers as `luria lint` warnings and the `luria reports` artifact. The
+    choice — reported, never enforced — is unchanged.
 summary: >-
-  Two standing reports — `luria ref-status` (references to retired documents)
-  and `luria pending` (undecided decisions by age AND citation count) — are
+  Two standing reports — reference status (references to retired documents)
+  and pending decisions (undecided, by age AND citation count) — are
   WARNINGS that never fail a build, with one summary line each from `luria
```
dmarx, line 24: "Never" is extremely strong. This ADR's assertion precludes promoting a mechanism to a guarantee, which runs counter to the philosophy of the library.

It's fine if we *default* to not failing the build, but if downstream users want that behavior, we should make it easy for them to activate/implement it. The language of this ADR acts as a soft gate pushing against that, and should be modified (i.e. this ADR should probably be superseded or at least renamed)

luria/cli.py
dmarx: Agreed, and done — the `RETIRED` table is gone in the latest push. Removed names fall through to ordinary `unknown command` + usage (which already lists everything that took the old jobs). ADR-030 now records the shim as a rejected alternative with your argument, and the devlog keeps the first draft as the failed approach: the instinct to preserve was the exact reflex the PR was cutting, pointed at itself. Propagated up the stack to #36 and #39.

---

luria/cli.py
dmarx: Removed entirely in the latest push, which I think moots both versions of the note: `cli.py` neither defines annotations that need deferring nor introspects any, so it needs neither the future import nor `annotationlib`. One flag on the PEP 749 route for elsewhere in the package: `annotationlib` lands in 3.14, and `pyproject.toml` currently floors at `>=3.11` — the other modules' `from __future__ import annotations` lines are load-bearing until either the floor moves or they provably don't need deferred evaluation.

---

luria/cli.py
dmarx: I looked at this and I don't think Fire pays for itself under the current delegation model, though it could under a different one — your call on which model you want:

- Naive `fire.Fire({"lint": lint.main, ...})` fights the existing design: every module parses `sys.argv` itself via argparse, so Fire would try to bind `luria journal new "title"` onto `journal.main()`'s (empty) signature before argparse ever runs. And Fire treats return values as output to print rather than exit codes, which breaks `luria lint` as a CI gate unless every entry is wrapped in a `SystemExit` shim — at which point the dispatch table still exists and we've added a dependency to keep it.
- Idiomatic Fire is the real alternative: delete the per-module argparse layers and expose typed functions (`lint()`, `index(check=False)`, `journal.new(title)`) that Fire drives directly. That genuinely deletes more code than it adds, but it's a rewrite of all eight commands' flag/help/exit-code surfaces, plus the first third-party runtime dependency beyond pyyaml — and it costs the vendor-one-file property ADR-030 kept.

The dispatcher left after this PR is ~40 lines with zero deps, so my lean is to keep it — but if you want the idiomatic-Fire shape, say the word and I'll file it as an issue and do the rewrite as its own PR (it deserves its own diff, not a corner of this one).

---

record/decisions.d/ADR-007.md
dmarx: Agreed — "never able to fail a build" hard-codes a policy where the library should offer a dial, and it does read as a gate against exactly the prose → convention → mechanism → guarantee promotion (DP-5) the project is built on. Since the choice itself changes (from "warnings, constitutionally" to "warnings by default, enforceable by configuration"), supersession is the right remedy rather than another in-place correction.

I'll do it the way you directed for #37/#38: file an issue and pair the superseding ADR with the actual mechanism (an opt-in config knob that promotes chosen warning classes to lint failures), so the new decision records a working dial rather than a reworded sentiment. Follow-up PR on the stack shortly.

---

record/decisions.d/ADR-030.md
```diff
@@ -0,0 +1,115 @@
...
+Meanwhile two of the survivors are real but not human-facing: `luria collect
+--commit` is invoked exactly once, in ci.yml — running it by hand consumes
+fragments a reviewer was meant to see on the branch — and `luria reports`
+exists to produce the artifact `actions/lint` uploads.
+
+## Decision
+
+**Two tiers, and the help text says which is which.**
+
+1. **Contributor commands** — `lint`, `link`, `index`, `journal`, `remotes`,
```
dmarx, line 61: 'filing an entry' currently only refers to the devlog, but we are essentially templating decisions and principles already as well. any scheme the user might want to associate with the repo should be discoverable in the config, so really any prefix-coded scheme could have its fragments templated and new entries mediated by the CLI. in service of that, I propose that rather than `luria journal`, what we actually want is something more like `luria new -` where `-` could default to the journal, but we'd also permit the user to do e.g.  `luria new adr` or `luria new dp` or `luria new changelog` etc.

Because documents all have frontmatters with multiple fields, I think rather than taking input from the command line to populate the document with (like `journal` does at present), `new` should just generate a new document and return the path to the user. `new` can automate constructing the filename and created time stamp. we can expose the option for the user to specify everything on the commandline by passing in named arguments for the different fields,  which might be useful for an LLM, but as a human user I want to be able to author my fragments in a markdown-aware editor rather than operating around an expectation that I'll be completing that data on the commandline when I definitely will only be using the CLI to create the document and then fill it out somewhere else afterwards.

record/decisions.d/ADR-030.md
```diff
@@ -0,0 +1,115 @@
...
+   warnings ([ADR-007](ADR-007.md)) and land in full detail in the `luria reports`
+   artifact.
+
+**A retired name answers with where the job went.** `luria badges` exits 2
+with "`luria index` writes the README's badges", not "unknown command" —
+DP-1 applied to the interface's own history. The pointers live in a
+`RETIRED` table in `cli.py`, beside the commands they replaced.
+
+**The modules keep their `main()`s.** Retiring a command removes a dict entry,
+not code: `python -m luria.ref_status --all` remains the interactive escape
```
dmarx: don't need to state this as if it's a rule. retiring a command absolutely can remove code and probably often should. removing code is desirable, it makes the code we are actually interested in easier to find if there's less sandbag code being dragged around.

record/decisions.d/ADR-030.md
```diff
@@ -0,0 +1,115 @@
...
+  documentation: eleven top-level names claim eleven workflows, and three of
+  those claims were false. A surface that lies about what you're supposed to
+  run is the same drift the record machinery exists to prevent, on the
+  machinery itself.
+- **Fold `link` into `lint --fix`.** The familiar linter idiom, and the two
+  already share `doc_refs.py` so they cannot disagree. Rejected as a separate
+  question: the check/fix split is a deliberate [ADR-005](ADR-005.md) shape, `luria link`
+  has a path-scoped mode `lint` has no analogue for, and merging saves one
+  name while this decision is about removing false claims, not shortening the
+  list for its own sake.
```
dmarx: no, this change is definitely also about shortening the list for its own sake. Yes, the check/fix split is deliberate. Isn't that shape preserved by the `--fix` flag?

I think the correct question here is: can we come up with a use case where a user might want to run `luria link` where `luria lint --fix` would be inappropriate and/or do more than we want?

Reflecting on this more, I think the answer is "yes". We don't just lint links, we also lint e.g. dates, and I could reasonably imagine tweaking `--fix` to handle date repairs or other fixes.

So yes, I actually agree that we shouldn't get rid of `link` outright, but not for the reasons stated here.

record/decisions.d/ADR-030.md
```diff
@@ -0,0 +1,115 @@
...
+- **Silent removal — plain "unknown command" for retired names.** Rejected on
+  DP-1: the dispatcher knows exactly what was asked for and where the job
+  went, and an explicit request deserves an explicit refusal that names the
+  successor.
+
+## Consequences
+
+- `luria --help`, the README's command table, both CLAUDE.mds and the Makefile
+  now describe the same six-plus-two surface; the `ref-status` and `pending`
+  make targets are gone.
```
dmarx, line 105: repeating the same thing in three places is a code smell, as is any mention of `CLAUDE.md`.

also, why do we have a makefile here? I'm guessing this was copied over as a vestigial component of StrataG's CI? I feel like we should be able to retire the Makefile in favor of direct CLI invocation. We should revisit what purpose the Makefile serves here.

record/decisions.d/ADR-030.md
dmarx: Filed as #42 and implemented in #43 (top of the stack): `luria new [kind]` — journal by default, any configured scheme by prefix (`luria new adr` copies `_template.md` to the next free number, stamps the date, substitutes the code into the body heading), any fragment dir by name (`luria new changelog`, filename from the branch slug). The contract is exactly as you framed it: compute identity, never content — it prints the path and gets out of the editor's way; `--title`/`--status`/`--summary`/`--tags` exist for a tool driving the CLI, never as requirements. Kinds derive from `luria.toml`, so a new scheme scaffolds for free, and `luria journal` is removed rather than shimmed. Fired before trusted: #43's own fragment and devlog entry were created with it.

---

record/decisions.d/ADR-030.md
dmarx: Rewrote the bullet with your reasoning in place of mine (pushed, propagated up the stack): the check/fix shape would indeed survive as a flag and shortening the list is a legitimate goal here, so neither is the objection — `link` earns its name because a future `lint --fix` plausibly repairs more than references (dates, populated fields, whatever else becomes mechanically fixable), and `luria link` is then the tool that does exactly the reference rewrite and nothing else. That future got a step closer in this stack, incidentally: #36 already has `luria index` repairing `created:` fields, so "fix" is provably a wider category than "link".

---

record/decisions.d/ADR-030.md
dmarx: Both smells filed as #45, with one datum that sharpens the Makefile half: since ADR-029 moved the docs jobs into the composite actions, CI's entire Makefile usage is one line — `make test` in ci.yml, itself a wrapper for `python -m pytest tests -q`. Every other target restates a CLI one-liner, and the `ref-status`/`pending` targets already had to be chased in this PR when their commands died — the drift is observed, not hypothetical. The surface-restatement half (help text, README table, CLAUDE.md block, Makefile help — four hand-maintained copies of `COMMANDS`) is in the same issue, since retiring the Makefile only deletes one of the four. Left as an open question per your framing rather than pre-decided; say the word and it becomes the next branch on the stack.

---

template/CLAUDE.md
```diff
@@ -50,11 +50,13 @@ luria lint            # the only one that can fail
 luria link --fix      # rewrite bare references as hyperlinks
```
dmarx, line 50: CLAUDE.md should be truncated to a short list of links to useful docs and invitation to run `luria --help` to learn what the package's current API is.

CLAUDE.md
```diff
@@ -59,11 +59,13 @@ luria lint            # the only one that can fail
 luria link --fix      # rewrite bare references as hyperlinks
```
dmarx, line 59: CLAUDE.md should be truncated to a short list of links to useful docs and invitation to run luria --help to learn what the package's current API is.

template/CLAUDE.md
dmarx: oh, this is in the template. What happens if we run `luria init` in a directory that already has a `CLAUDE.md`? presumably that file isn't modified and luria raises a warning or something like that, right? Opinionated as I may be, we probably shouldn't be overwriting people's claude.md files w/o their permission. Maybe we could append a message to the bottom with instructions to the LLM recommending it simplify its CLAUDE.md per our idiom.

Makefile
```diff
@@ -26,13 +26,7 @@ index: ## regenerate the decision index from frontmatter
...
 	luria journal
 
-ref-status: ## references to retired documents, every site
-	luria ref-status --all
-
-pending: ## undecided decisions by age and citation count
-	luria pending
-
-reports: ## write both reports to build/doc-reports/
+reports: ## write both status reports to build/doc-reports/
```
dmarx, line 29: yeah it suure looks like we probably don't need this makefile at all.

README.md
```diff
@@ -173,14 +173,17 @@ luria index && luria lint
 | `luria lint` | the only command that can fail: index completeness, frontmatter, a stale generated index, and references that should be links |
```
dmarx, line 173: would probably be more user friendly if we explained the available commands in the form of a normal workflow rather than essentially echoing the --help message.

template/CLAUDE.md
dmarx: To the factual question: `luria init` never overwrites — an existing file prints `skip CLAUDE.md (exists)` and lands in the "skipped N existing" count, so nobody's CLAUDE.md is at risk.

Both halves of the rest are done in #46: the truncation (both CLAUDE.mds become maps — links + `luria --help` + three one-line ground rules, ADR-037), and the append idea landed as the gentler variant — when init skips an existing CLAUDE.md it now prints a tailored pointer at the map shape and suggests asking your agent to fold it in, on **stdout** rather than appended into the file. Writing into a file luria doesn't own seemed like the overwrite problem in miniature; a printed recommendation needs nobody's permission. Easy to flip to append-with-marker if you'd rather it land in the file.

---

CLAUDE.md
dmarx: Done in #46 (ADR-037): both CLAUDE.mds are now maps — a short link list to the authoritative docs, the invitation to run `luria --help`, three one-line ground rules, and the statement that when the file disagrees with the docs or the CLI, the file is the one that's wrong. The command block had drifted twice in one week, so the evidence was already in.

---

Makefile
dmarx: Gone, in #46 (ADR-038): ci.yml runs `python -m pytest tests -q` directly, the Makefile leaves the scan globs, and `luria --help` is the one list of what you can run. Deleting it also surfaced a straggler worth the trip — `luria.toml`'s devlog blurb still said `luria journal new`, hiding in a `.toml` string my earlier sweeps didn't cover.

---

README.md
dmarx: Rewritten in #46: the table became a three-beat workflow narrative — *file it while it's warm* (`luria new …`), *let the machinery write the plumbing* (`link --fix`, `index`), *check before you push* (`lint`, with the `fail_on` dial noted) — with the occasional/CI commands in a closing paragraph and the same last line as the CLAUDE.md map: `luria --help` is the authoritative list. A table that echoes `--help` is a copy; a narrative that shows *when* you'd reach for each command is content the help text can't carry.

---

luria/cli.py
dmarx: > Idiomatic Fire is the real alternative

Yeah that's what I was proposing. implement it as a draft PR for after we merge this stack and then writeup your thoughts on the PR regarding whether or not the refactor was worth it. I think it will be.

dmarx: Drafted as #48 (top of the stack), with the worth-it writeup in the PR body. Short version: ~150 lines of argparse deleted across ten modules, flags now derived from the typed `run()` signatures so the parser can't drift from the function it wraps; the honest costs are the first runtime dependency beyond PyYAML and one piece of discipline the old shape enforced by type signature — under Fire, *returning* a status prints it and exits 0, so failure must be `SystemExit` (a silent CI-gate bypass if a future command forgets). ADR-039 rides along as `Proposed` — the merge verdict is the decision, and the `needs decision` badge going orange over it is the dogfood working. My verdict, in full in the PR: narrowly worth it, for the drift-deletion rather than the line count.

---

PR #48, opened 2026-08-07
DRAFT: drive the CLI with Fire — typed functions, derived flags (ADR-039)

dmarx:
The idiomatic-Fire rewrite from the #32 review thread, as requested: a draft for after the stack merges, judged on the diff. Top of the stack (#46 → #43 → #41 → #39 → #36 → ~~#32~~ merged). ADR-039 is filed as **Proposed** — this PR is the question, the merge verdict is the decision, and the `needs decision` badge going orange over it is the dogfood working.

## What changed

- Every command is a plain typed function (`<module>.run`); `fire.Fire(COMMANDS)` derives flags and help from the signatures and docstrings. The hand-rolled dispatcher and **every module's argparse layer are deleted** (~150 lines across ten modules, replaced by nothing).
- **Exit codes are SystemExit only.** Fire prints return values, so `lint` returning 1 would print "1" and exit 0 — a CI gate that green-lights everything. Every `run()` returns `None` and raises to fail.
- Modules stay standalone (`python -m luria.ref_status --all`) via `fire.Fire(run)` under their `__main__` guards — the vendor-one-file story survives with one convention instead of two.
- **The invocation surface is unchanged**: every spelling CI and the composite actions use (`lint`, `link --fix`, `index --check`, `new adr --title X`, `collect --commit`, `init --issue-url U`) parses identically. Smoked end-to-end (init → index → new adr/dp → lint) in a scratch project; exit codes byte-identical. 296 tests, lint clean.

## The writeup: was it worth it?

**What it bought, concretely:**
- ~150 lines of argparse deleted, and — the real prize — **a flag surface that cannot drift from the functions it calls**, because it *is* the functions. The old shape had five hand-maintained restatements of the surface; this removes the deepest one (each parser restated its own function's parameters).
- One convention for everything: the CLI, `python -m`, and the tests all drive the same typed `run()` functions. Adding a command is now: write a typed function, add one dict entry.
- Unknown commands get "Cannot find key: X" plus the available-commands list — the ADR-030 refusal property, free.

**What it cost, honestly:**
- The first runtime dependency beyond PyYAML, for a job 40 zero-dep lines were doing.
- **The exit-code rule is now discipline instead of a type signature.** The old `main() -> int` shape made "return a status" the obvious thing to do and the correct thing to do; under Fire, returning a status is the obvious thing to do and a *silent CI-gate bypass*. This is the strongest argument against, and it's recorded in ADR-039's consequences so the next command author meets it.
- Help output is Fire's house `NAME`/`SYNOPSIS`/`FLAGS` format — correct and derived, but the tiered contributor/CI usage text is gone (the tiering now lives in the docs), and Fire's error footer is chattier than the old one-line refusal.
- One trap found during smoke testing, recorded in the devlog: `luria frobnicate; echo $? | tail` measures the pipeline, not luria — Fire briefly looked like it swallowed exit codes and didn't.

**My verdict:** narrowly worth it, and for a different reason than line count — the deletion that matters is the *parser-restates-the-signature* duplication, which was this codebase's own DP-3 drift shape living inside every module. The dependency and the exit-code discipline are real prices; the derived-from-code surface is the kind of guarantee this project builds monuments to. If the exit-code discipline worries you more than the drift did, the honest alternative is closing this unmerged and flipping ADR-039 to `Rejected` — which the record is shaped to do gracefully.

---
_Generated by [Claude Code](https://claude.ai/code/session_018AHHTjpaGcX2eQfE9kgvkd)_


dmarx:
> Fire prints return values, so lint returning 1 would print "1" and exit 0 — a CI gate that green-lights everything. Every run() returns None and raises to fail.

instead of returning the integer 1, couldn't we just invoke `sys.exit(1)` instead? we can still pass the return code to the OS, we just have to do  it through the sys module rather than as a return value. we aren't bound to POSIX rules here so it's not like there's an expectation anywhere that our tooling returns exit codes.

dmarx:
> couldn't we just invoke `sys.exit(1)` instead?

Yes — and that's exactly what shipped. Every `run()` raises `SystemExit(n)` on failure (`sys.exit(n)` is literally `raise SystemExit(n)`; same mechanism), and it propagates cleanly through Fire to the OS — `luria lint` exits 1, `index --check` exits 1, both pinned by subprocess tests. Exit codes are fully preserved.

The PR body's note is only about the *trap for future commands*: under argparse the natural idiom (`return 1` from `main() -> int`) was also the correct one, while under Fire that same natural idiom silently becomes "print `1`, exit 0". So the rule is one line — raise, never return a status — it's just now a convention ADR-039 states rather than a shape the old `-> int` signature enforced.

Also: the rest of the stack is merged (as #49–#53 — details in my next status comment), this branch is reconciled with main, and the PR is retargeted, so the diff you're reviewing is Fire-only.

---

Issue #55, opened 2026-08-07
Execute the DP→GP rename as the first live migration


dmarx:
The execution half of #29, running on the machinery from #54. Blocked on #54 rung 2.

The rename itself, decided in #29: **design principles → guiding principles**, `DP-00N` → `GP-00N`. Three-letter-adjacent prefix retained on purpose; "DP" was always a slightly cursed abbreviation. The companion "architecture → atomic decision records" relabel is rung 0 — prose only, no codes move — and rides along as the worked example of the cheapest rung.

Spec: `record/migrations.d/0001-dp-to-gp.toml` (scaffolded by `luria new migration`), a single `rename_scheme` operation. What the run must produce:

- `[luria.schemes.DP]` → `[luria.schemes.GP]` in `luria.toml` and `template/luria.toml`; output becomes `docs/guiding-principles.md`; the `[luria.paths]` key follows.
- `git mv record/principles.d/DP-00N.md → GP-00N.md` (9 documents), each stamped `formerly: [DP-N]`.
- Mapping-driven sweep of every reference: link targets, labels, and anchors (`#dp-N` → `#gp-N`), historical journal and changelog files included — full rewrite per the ADR-040 doctrine, one spelling tree-wide.
- **Not touched:** fixture `DP-018` in `tests/test_remotes.py`, example `DP-004` in `remotes.py` (both carry `unresolved-ok` excuses), and `SG-DP-18` / `[luria.remotes.SG.schemes.DP]` — strata-g's namespace is theirs.
- Migration commit appended to `.git-blame-ignore-revs` by the tooling.
- `luria lint` green with zero `legacy-spellings` warnings in-tree afterwards; branches in flight get modernized by `luria link --fix` on their next touch.

This run is the acceptance test for the machinery: if the DP→GP migration needs hand-editing beyond writing the spec file, rung 2 isn't done.

PR #61, opened 2026-08-08
The DP→GP migration: rung 2 machinery + the run (rebuilds on #62)

dmarx:
**Draft, verdict-shaped (ADR-042). Blocked on #62**, which now carries everything from this workstream worth keeping regardless of the verdict here — the status flips, ADR-042 itself, rung 1, the fixture-doctrine cleanup, two of the three stamp hardenings, and DP-011. That factoring is ADR-042's rule 3 in action: rejection stays cheap, and this draft now stakes only what the verdict is actually about.

**What this PR is, after #62 merges and this branch is rebuilt on the new main:**

- **Rung 2 machinery**: `luria migrate` executing specs from `record/migrations.d/` (`rename_scheme`, `move_doc`, `strategy="supersede"`, `--dry-run`, `--commit` with `.git-blame-ignore-revs` bookkeeping), `luria new migration`, and the sweep with its masks — composed foreign codes, URLs on path pairs, `formerly:` blocks, `unlinted-file` pages, section-aware config passes.
- **The run** (#55's acceptance test): the spec, re-executed fresh — which is the point: the run is reproducible from the spec, and the new mapping picks up DP-011 automatically. First execution's ledger, for the verdict: 10 documents moved and stamped, 699 rewrites in 117 files, machinery alone did config edits/moves/stamps/sweep/view relocation/blame bookkeeping; residue was the template's seed files (downstream project — rung 3, deferred), the rung-0 prose relabel (deliberately human), and exactly one string no pattern could claim (a bare `"dp-4"` dict literal in a test assert).
- The run devlog entry and the migration changelog fragment.

**The verdict weighs** (full writeup preserved from the first run): *for* — the acceptance test substantially passed, three guards fired live and now hold structurally; *against* — ~470 lines of rarely-run machinery, a mask list that encodes judgment, and a knowably-open pattern set (an anchor quoted as a bare value is invisible to it).

Closing files ADR-042's first use as a rejected experiment and the record keeps rung 1 and everything else via #62.

---


dmarx:
Closing — the machinery this PR was blocked on shipping has shipped, and nothing here is left that `main` lacks.

Measured rather than assumed:

```
git diff --name-only main pr61 -- luria/ tests/ | while read f; do [ -f "$f" ] || echo "ONLY-IN-61: $f"; done
(no output)
```

Every module and test this branch introduced is on `main`, via #80 (the machinery, ported without the run) and #84 (provisional-code allocation, the `Pair` split, and a concretize anchor fix). The branch's own diff-vs-main now reads as ~2,400 deletions, because it forks before five feature PRs.

**What remains is the DP→GP run itself** — the spec plus the rename of this record — which is the verdict-shaped half you pinned, and which the machinery deliberately shipped without. That is one 20-line spec, regenerable with `luria new migration` whenever the verdict is made; the PR body's writeup of the trade survives here in the closed thread and in the devlog.

Two things learned since, worth carrying into any future DP→GP run:

- **A `rename_scheme` is unaffected by #86** (`move_doc` across differing `render` modes leaves dead links), since both sides of a rename share a render mode. So the run is still safe to execute as specified.
- **`allocate = "merge"` is now set on the `DP` scheme**, so a re-cut spec inherits provisional codes rather than claiming numbers.

Reopen if you'd rather keep the branch as the place the run lives; nothing in it is lost by closing, since the spec is derivable and the machinery is merged.

---

PR #66, opened 2026-08-08
DP: exempting a ledger from one matcher exempts it from none of the others

dmarx:
Recut per review. The observation and the evidence were right; the document was carrying two claims and a sustained metaphor.

## What changed

**It was two principles.** That a ledger is spelled like the thing it records, and that exemptions should be structural rather than incidental. The title needed a colon to hold both — and the second is the weaker half, close to generic advice and already partly covered by DP-6, which the body cited.

**The sharper claim was in the evidence**, and this PR's own workshop questions named it: *"every hunter needs its own mask, because masks don't transfer."* That is what three independent discoveries actually prove.

> **Exempting a ledger from one matcher exempts it from none of the others**

Each subsystem matched the pattern on its own terms, so fixing the first protected nothing else. An exemption is a property of the *matcher*, not of the ledger, and there is nowhere to put one where every matcher will see it. That version is predictive — it says what to do the next time a consumer of a pattern is added — where "be structural, not incidental" is advice.

**The metaphor is gone.** Prey, hunters and mouths ran through every paragraph, and *"three mouths, three separate discoveries"* read as decoration over the finding. The siblings state their claims plainly (*Culture must be compiled*, *Fire before trusting*), and DP-1's one metaphor is a clean equation rather than a sustained conceit.

**The hypothetical examples are gone**, replaced by a real second instance. A redirect map and a suppression list never happened here; this did, and it is a subsystem that got it right:

> The reference scan blanks the span of any acknowledgement before counting citations, because `inactive-ok: ADR-012` names the very code the retired-citation check looks for — without the mask an annotation would excuse itself, and could never go stale. It matches the directive's **shape** rather than the parsed directives, so an example of one in the documentation is covered too.

That is structural-not-incidental demonstrated instead of asserted, and it is three masks in one subsystem — which is the transfer claim again, from the other direction.

## Answering the workshop questions

- **Is "structurally, not incidentally" the right cut?** No — "masks don't transfer" is, as you suspected. The structural/incidental distinction survives as a paragraph rather than as the thesis.
- **Human-process reading?** Left out. It dilutes a claim whose whole force is that independent matchers each need their own mask, and there is no evidence for it here.
- **`influenced_by` is thin.** Gains ADR-049, whose `aka:` aliases are the other ledger of this kind in the record.

## Also

Filed `Active` rather than `Proposed` — the verdict this PR asked for is in, and leaving it `Proposed` would reopen a pending queue that currently reads zero. Merged `main` in; the only conflicts were generated views.

541 tests pass; `luria lint` clean. Register pass run over the document: four flags down to two, both pre-existing shapes in the origin field and the em-dash density, which sits at 1.3 against this repository's 2.5–3.5.


dmarx:
Rebuilt on current `main` under a **provisional code** — `DP-tmpqu8fy` — rather than claiming `DP-011`.

The branch forked before merge allocation existed and took `DP-011` by hand. That number has since been assigned to a different principle on `main`, so this PR was carrying a live collision: the exact failure [ADR-049](record/decisions.d/ADR-049.md) exists to prevent, and one that would have silently overwritten a document at merge. `luria concretize` now assigns the real number at the serialization point.

**Recreated rather than rebased.** The fork point predates roughly 10,600 lines of subsequent work — `migrate.py`, `concretize.py`, `site.py`, `narrow_titles.py`, the `examples/` tree and a dozen ADRs all postdate it — so a rebase would have been a fight with deletions rather than a re-application of the one document this PR is about. The principle's text is unchanged; only its own code is re-spelled, and every other mention inside the file is a self-citation that moves with it.

The changelog fragment drops the hardcoded number too, since it no longer knows one.

413 tests pass; `luria lint` clean.

Worth noting the principle now has a fourth worked instance it didn't have when drafted: `luria concretize`'s sweep was rewriting temporary codes but walking straight past the *anchor* spelling of the same code, because the replace was case-sensitive — a hunter that missed part of its own quarry rather than eating its ledger, but the same class of "the pattern is spelled more than one way and the mechanism only knows one". Fixed in #84.

---

Issue #57, opened 2026-08-08
Journal links live in the book's frame: correct-at-source links break the rendered view, and lint doesn't notice


dmarx:
Found while filing the devlog entry for ADR-040 (PR #56): a hand-written relative link broke the generated monthly book, and only the package's own pytest caught it — `luria lint` stayed green. Characterized below with a minimal working example.

## Mechanism

A journal entry's body is transplanted **verbatim** into its book — `render_book` (`luria/journal.py`) does no link rebasing — so a relative link can be correct in exactly one of two frames:

- the **source frame**: the entry's own directory, `record/devlog.d/YYYY/MM/DD/` (what GitHub's file view resolves against);
- the **book frame**: the render destination, `docs/devlog/` (what the published view resolves against).

The project chose the book frame: `link_base()` (`luria/doc_refs.py`, "Link bases" section) maps every journal entry to `docs/devlog/`, and `luria link --fix` writes targets accordingly. That choice is coherent, but it has three sharp edges:

1. **Fixer-written links are broken at source.** Browse any devlog entry file on GitHub and its ADR links 404. Working as designed, but nothing tells the reader that.
2. **A source-relative link — the natural spelling, and the one GitHub's file view rewards — silently breaks the book.** No lint check resolves relative link targets. The property is guarded solely by `tests/test_adr_index.py::test_every_generated_relative_link_resolves`, which exists only in this repo's test suite. **An adopting project has no pytest: green lint, broken published view.**
3. The convention is documented only in a module docstring, which no entry author reads.

## Minimal working example

```console
$ mkdir mwe && cd mwe
$ luria init --issue-url "https://example.com/issues"
$ luria new adr --title "A decision to cite"      # ADR-001
$ luria new --title "An entry citing a decision"  # record/devlog.d/2026/08/08/001620.md
```

Write a source-relative link into the entry body — correct from the entry's own directory:

```markdown
[the numbering decision](../../../../decisions.d/ADR-001.md)
```

```console
$ luria index && luria lint; echo "lint exit: $?"
luria: docs lint clean
lint exit: 0
```

But the same string, verbatim in `docs/devlog/2026-08.md`:

```
link as written:             ../../../../decisions.d/ADR-001.md
resolves from the entry:     True
resolves from the book:      False      ← published view is broken; lint was green
```

The dual — replace the body with a bare `ADR-001` and run `luria link --fix`:

```
what the fixer wrote:        ../../record/decisions.d/ADR-001.md
[...]

PR #75, opened 2026-08-11
Generate the configuration reference, and say that the shipped four layers are a default

dmarx:
Luria supports far more than it lets on, and the reason turned out to be interesting: **the generality was never undocumented — it was misfiled.**

`config.py`'s dataclass docstrings are in practice a complete configuration reference. `template/luria.toml`'s comments are good and say things like "Add another (RFC, SPEC) as a third entry; nothing else needs to change". ADR-006, ADR-024, ADR-028 and ADR-036 cover schemes, uid remotes, collection styles and config-driven scaffolding. Three good explanations, none of them anywhere a reader arrives in time: docstrings need you to know which module to open, the template is seen *after* `luria init`, and the decision record is archaeology — it answers "why", which is the question you ask second.

`docs/` had no configuration page at all, and no mention of `luria.toml` outside generated ADR summaries. So the shipped four layers read as Luria's *parts* rather than its *defaults*, which suppresses the question whose answer has always been yes.

## The capability surface, as found

| primitive | table | generalized capability |
|---|---|---|
| Scheme | `[luria.schemes.X]` | any referable document family — prefix, dir, in-force status, `render = index\|document`. Reads `ADR-013.md` *and* `adr-13-a-slug.md` |
| Fragment | `[luria.fragments."dir"]` | any many-files→one-document pipeline; `style = append\|changelog` |
| Journal | `[luria.journals.X]` | any timestamped, never-consumed entry stream → books. It's a dict, so **multiple journals work** |
| Remote | `[luria.remotes.X]` | cross-reference resolver — and with `uid` + `url`, a general citation linter for arXiv IDs, ticket keys, CVEs |
| Site | `[luria.site]` | publish any of it as a Quartz vault, branded, all derived from `issue_url` |

## The change

**A generated configuration reference.** Writing the missing page by hand was the obvious move and the wrong one — it is the second copy of a schema that changes, the projection DP-3 exists to forbid, and it would drift specifically for newly added keys, which are the ones a reader most needs. So `luria index` renders `docs/configuration.md`: prose from each config dataclass's docstring, key tables from `dataclasses.fields()`.

The load-bearing property is that the key tables are *read, never listed* — a key that exists in the schema is a documented row whether or not anyone remembered to describe it. Where classification is hand-held (which keys Luria fills from the table name, which derive from `issue_url`) it is presentation only, and an unclassified field still gets a row, merely labelled `*required*`. That is the safe direction to be wrong in.

Registering the page in `Config.is_generated` does double duty: staleness is checked like any other view, and `doc_refs.doc_files` filters on the same flag, so a page built out of example codes is skipped by the bare-reference lint and left alone by `luria link --fix` — the pattern the status reports already established.

**`docs/adopting.md` gains "Shaping the record to your project"**, before the migration chores so a reader meets the dials first: a second document family, index vs document rendering, keeping a collocated layout, a second journal, fragment styles, `uid` remotes, and the `fail_on` dial.

**The README keeps its decision-record framing** and gains one paragraph plus a docs-list entry. The scaffolded `CLAUDE.md` points adopters' agents at the reference.

All of it states a limit rather than leaving it to be discovered: adding a scheme costs one table, but **renaming** one is still a manual pass — ADR-040 is the decision that would give it a command.

## Verification

- `python -m pytest tests -q` → **349 passed** (331 before, 18 new)
- `luria lint` → clean; `luria index --check` → current and idempotent
- `luria site` stages the new page (62 pages, links resolve)
- Verified rather than trusted: `python -m luria.ref_status --all` actually runs, since the reference now publishes it as a capability

**Both guards fired once before being trusted (DP-6)**, recorded in the devlog:
- Hand-editing the page makes `luria index --check` exit 1 and name it.
- A throwaway field added to `Site` became a documented row **with no edit to the generator**, and made the committed view stale — the whole design demonstrated in one command.
- Unprompted bonus: the docs-index check caught the missing `docs/README.md` entry on the first run after the page appeared.

## One gap found, deliberately not fixed here

`luria link --fix` does not detect bare `DP-N` references, though `CLAUDE.md` and the scaffolded template both promise it does. `find_refs` returns nothing for `DP-3` anywhere tested, while `ADR-012` and `#57` resolve normally; the anchors exist and are correct, and the wikilink form `[[DP-3]]` expands perfectly (which is what this contribution's prose uses).

So the affordance works and the documented shorthand does not, **silently** — a bare `DP-6` is neither linked nor reported, the worst of the three possible behaviours. That is a change to shared reference machinery and deserves its own contribution rather than a ride-along in a docs change, so it is written up in the devlog instead. Happy to open it as a follow-up.

## A note on ADR-044's status

Filed `Proposed`, not `Active` — the decision is the reviewer's to make, and merging is what makes it true.

---


dmarx:
## Update: the configurations are now executable, and running them falsified two of my own claims

Pushed three more commits. The docs in the original push were still prose — accurate as far as anyone had checked, and unchecked. So `examples/` now holds four complete projects, and `tests/test_examples.py` builds each into a temp tree and runs the real `luria index` and `luria lint` against it.

| example | demonstrates |
|---|---|
| `rfcs-and-specs/` | two families side by side, one `render = "index"`, one `render = "document"` |
| `collocated/` | views beside sources — adoption without moving files |
| `many-journals/` | three journals at month / year / day granularity |
| `external-citations/` | `uid` remotes: arXiv, Jira (custom `delim`), CVE — none of them a record |

Generated views are deliberately **not** committed. The tests build in `tmp_path` and `examples/.gitignore` keeps output out of the repo, because a committed view no CI run regenerates would be a stale projection sitting in the one directory people copy from. `examples/**` joins `template/**` in the site exclusions.

### The first pass disproved two things I had written that same day

Both the same shape — true of the mechanism, false of the shipped defaults, and invisible to review:

1. **`active = "Accepted"` was wrong.** The five statuses are closed and lint-enforced (ADR-003). `active` *selects* from the vocabulary; it cannot extend it. As written, the example named a state no document can hold and would have failed every document in the scheme.

2. **"Omit `output` and the view renders beside its sources" was wrong for `ADR`** — the sharp one, because it hits the documented adoption path. Config merges over `DEFAULTS`, which carries `output = "docs/decisions"`. A project pointing `dir` at its existing decisions and omitting `output`, expecting to keep its layout, gets its index silently relocated. The working form is `output` equal to `dir`.

A third defect turned up too: `DEFAULT_DOCUMENT_STUB` hardcoded `# Design principles`, so a SPEC family rendered as a document came out titled after this package's principles. Fixed — the fallback names its own scheme.

### One regression I caused, and chased

Rendering the config reference for *every* project meant every project suddenly needed a `docs/configuration.md` index entry. A fresh `luria init` failed `luria lint` on the first run — two lines below the adoption guide's promise that it would be clean. Fixed in `template/docs/README.md`. Existing projects upgrading will see one violation naming the missing entry exactly.

Caught by scaffolding a throwaway project and running the documented three commands. The guide is a script; worth running it.

### Left as documented limits, not fixed

Two of the three findings share a cause: **configuration merges over `DEFAULTS` rather than replacing them.** So the shipped `ADR` scheme cannot be removed, and its keys cannot be unset by omission — only overridden. Changing merge semantics is its own decision, and making it quietly inside a docs contribution is the silent revision this record exists to prevent. The reference, the guide and `examples/README.md` all say so, and three tests pin the limits so a documented constraint can't drift either.

Happy to open that as a follow-up ADR if you want it resolved rather than described.

**365 tests pass** (331 at branch point), `luria lint` clean, `luria index --check` current.

ADR-045 is filed `Proposed` alongside ADR-044 — both are yours to decide.

---

dmarx:
## The `DP-N` gap was the visible corner of a much bigger one

I went in to fix bare `DP-6` detection. The actual defect is that **reference detection was never scheme-driven at all.**

```python
for kind, regex in (("dp", DP_RE), ("adr", ADR_RE), ("issue", ISSUE_RE)):
```

Three hardcoded patterns. A configured scheme's own `Scheme.pattern` — which exists, and which `luria new` and every renderer already use — was never consulted for finding references in prose.

So a project that configures `RFC` gets an index, tag pages, `luria new rfc`… **and no reference checking whatsoever.** `RFC-7` in prose is neither linked nor reported.

The proof was already sitting in the examples from the last push: `examples/rfcs-and-specs` had an RFC citing `SPEC-001` in plain prose, and `luria lint` passed it. The package whose entire purpose is catching unlinked references wasn't catching them for any scheme but its own default.

ADR-006 made schemes configuration rather than constants. That generality reached rendering and scaffolding and **stopped one layer short of the linter — the layer the promise was about.**

### The fix

`find_refs` builds its patterns from `current().schemes` and tags each match with its prefix. `Ref.kind` collapses to `scheme | issue | remote` — a per-scheme kind would make `ADR` special again, which is precisely the vocabulary ADR-006 rejected when it chose `inactive-ok` over `adr-ok`. `resolve` dispatches on render mode: a file link for `index`, an anchor in the assembled page for `document`.

`DP_RE` still runs first for the prose spelling (`design principles #17` and its sibling chain), which no scheme pattern covers. Remotes still claim their spans first, so `LU-ADR-013` is never read as a local `ADR-013`.

Cross-scheme references now resolve both directions, each from the base where its text renders:

```
RFC-001 → SPEC-001   ../../docs/interfaces.md#spec-1
SPEC-001 → RFC-001   ../record/rfcs.d/RFC-001.md
```

### Blast radius

**38 references in this repo linked in one pass** — all `DP-N` codes that had accumulated while the lint said clean. Worth reading a sample of that diff.

One guard was load-bearing and not obvious: **a document never links to itself.** For an index scheme that falls out of `target == source`. For a document scheme the source is the *fragment* and the target is the page it assembles into — different files, so that test never fires. The first run turned every principle's own `# DP-001:` heading into a link, which would then have failed `check_title`. Caught by reading a sample of the diff, which is advice already in `docs/adopting.md`.

Two illustrative DP codes in `doc_refs.py`'s own prose started resolving to nothing once visible, and needed an `unresolved-ok` directive. Not a regression — wrong before too, silently. **That's the shape of the upgrade every adopter will see**, so the changelog says so explicitly.

### I changed my mind about splitting this out

I said this deserved its own contribution rather than a ride-along. It landed here anyway: a repo-wide relink stacked on a branch that already rewrites the same files buys a merge conflict, not clarity. Recorded in the devlog rather than done quietly — happy to split it if you'd rather review it alone.

**369 tests pass** (331 at branch point), lint clean, index current. ADR-046 filed `Proposed` alongside 044 and 045.

---

dmarx:
merge it and then work on that other thing you pointed out. "the shipped ADR scheme cannot be removed, and its keys cannot be unset by omission — only overridden." -- can `luria init` not take a config file as input?

PR #77, opened 2026-08-12
Merge-allocated schemes: temporary codes, concretization, and permanent aliases (ADR-049)

dmarx:
Implements the proposal filed in #76 as ADR-049, from your brainstorm there: numbers are a global counter, and a branch is the wrong place to claim one.

## The loop

```toml
[luria.schemes.ADR]
allocate = "merge"          # default stays "filing" — nothing changes unless you opt in
```

```
luria new adr               → record/decisions.d/ADR-x47fje.md
# cite it freely on the branch: bare ADR-x47fje or [[ADR-x47fje]] — indexed, linted, linked
luria concretize            → ADR-x47fje → ADR-123, tree rewritten, alias recorded
luria concretize --check    → CI guard for main: exit 1 naming any temp code
```

## The design choices worth reviewing

**The temp shape is disjoint by construction.** Six base-36 chars, first forced alphabetic — so `number_of`/`temp_of` and the two prose patterns can coexist with no precedence rules. Every place that had to "handle both" became two patterns that cannot overlap.

**Aliases are the load-bearing half.** Concretization appends the temp code to the document's `aka:` frontmatter, and the resolver honors it forever, in both spellings. A temp code in a PR thread, an immutable commit message, or another repo's `LU-` reference survives as an old name, never a dead link — ADR-014's contract extended through the rename. Alias lookup runs only on the rare path (temp code with no live doc) and re-reads frontmatter each time: slow-but-rare beat a cache that would need invalidating at exactly the moment the concretizer renames files.

**Historical files are not rewritten.** Devlog entries and collected changelogs stay as written; the alias keeps their citations resolving. Rewriting history to keep it accurate would be the silent revision the record forbids.

**The tree rewrite is one string replace, on purpose.** A temp code's filename is the code plus `.md`, so `ADR-x47fje → ADR-123` rewrites link labels and targets in a single pass — no markdown parsing. Assignment order is commit time, the same ordering the changelog collector trusts, so the sequence records the order decisions entered the record — which is what sequential numbering already claimed to mean.

**Ordering caveat:** temp docs added in the *same commit* tie on commit time and fall back to name order. Distinct PRs (the actual use case) have distinct commit times.

## Found while firing it

- First hand-fired run crashed `render_categories` — it formats `number:03d` and a temp doc has no number. The integration break no unit test of the minter would have seen; the tests now drive the whole loop through a real record.
- Self-links nearly returned: `aka:` makes every document reachable under two names, so the guard fires on the *resolved target*, not the spelled code. Caught by the fixture citing in both directions.
- The template docs README's `[LU-ADR-048]` link label read as a local `ADR-048` citation in scaffolds that don't register the LU remote — reworded to a prose label.

## Scope line

This repo's own record **stays on filing allocation** — ADR-049 says the mechanism existing and this record adopting it are two decisions. Switching would want `allocate = "merge"` in `luria.toml` plus `concretize --check` wired into CI on main, in one change. Say the word.

The configuration reference picked up `allocate` **without being touched** — ADR-044's derivation guarantee doing its job.

## Verification

- **388 tests pass** (381 at branch point): 8 new driving mint → cite → lint → link → concretize → alias, plus the trunk guard and the filing-default pin
- Fired by hand first: the full loop on a scratch record, including cross-scheme citation, alias resolution from a third file, `--check` both sides, and idempotent re-run
- `luria lint` clean, `luria index --check` current
- ADR-049 remains `Proposed` — implemented is not the same as adopted, and the flip is yours

---


dmarx:
Re: marking the code as provisional (`72ea3f7`) — the question split into one idea worth taking and one worth rejecting, and no, I don't think it's convoluted; the legibility gap was real.

**Taken: a spelled-out sentinel in the code.** Temp codes are now `ADR-tmp47fje` — literal `tmp` plus five base-36 chars. Three things for the price of one:

1. **Legible.** A reader who has never met the convention sees "provisional" at every citation site. The previous shape (`ADR-hlfrcp`, any six alphanumerics starting with a letter) required knowing the rule.
2. **Precise.** The old prose pattern false-matched six-letter English after a prefix — *"the ADR-review process"* parsed as a temporary reference, because `review` is six letters. Naming the legibility problem found a matcher bug nobody had named. There's a test pinning it now.
3. **Honest forever.** The code survives as a permanent `aka:` alias, and `ADR-tmp47fje` in an old commit message reads exactly right about what it was at the time. Entropy is still 36⁵ ≈ 60M per scheme.

(No internal hyphen — `tmp` lives inside the tail, so nothing that splits codes on `-` changes.)

**Rejected, recorded in ADR-049's alternatives: a provisionality state in frontmatter.** The code's state is already fully determined by its filename shape — `temp_of()` reads it directly — so an `id_state:` field would be a second copy of a derivable fact, which is precisely the projection DP-3 says will drift: the concretizer renames the file and must also remember to flip the field, and the day it forgets, the two disagree. Document status isn't this either: `Proposed` is the *decision's* state; the *code's* state lives in its shape, where nothing has to keep it true.

**On "is this getting too convoluted":** the mechanism has exactly three moving parts — mint, concretize, alias — and each is load-bearing (collision-freedom, the sequence, permanence of citations). The sentinel *reduced* net complexity: one glance now tells a human what a code is, and the matcher got stricter, not looser. The frontmatter-state variant is where I'd have agreed it tipped over — that's the one that adds state to keep in sync.

389 tests pass, lint clean, index current.

---

dmarx:
Ruling applied — the sweep is now full, history included.

`concretize` no longer skips historical files: journals and the collected changelog are swept like everything else, so **exactly one spelling of each code exists in the tree after a run**. This adopts ADR-040's second commitment outright, on its own reasoning — a swept run is maximally non-silent (one named commit at the serialization point), so it isn't the quiet revision ADR-019 forbids, and git rather than the working tree is the custodian of what was actually written.

The `formerly:` alias's job narrows to what it was always uniquely good at — citations no rewrite can reach:

- PR threads and immutable commit messages
- other repositories' `LU-`-prefixed references
- **branches cut before concretization**, which merge clean, resolve through the alias, and modernize on their next `luria link --fix`

ADR-049's open-tension paragraph is now the record of the ruling (with your "temporary is relative to the historical record" as the stated principle), the leave-history-alone test is flipped to pin the sweep, and the devlog keeps the honest narrative that the first version got this wrong and why the reversal is sound.

389 tests pass, lint clean, index current.

---

Issue #14, opened 2026-08-04
some broken content in the ADR README


dmarx:
https://github.com/dmarx/luria/blob/main/docs/decisions/README.md

* ADR-003 shows status as "Proposed" instead of "Active", and summary field is truncated
* ADR-008 summary split and leaked into status field
* ADR-025 same issues as ADR-008

Issue #35, opened 2026-08-07
Badges should link to reports


dmarx:
Instead of only uploading reports of stale decisions as CI artifacts, let's actually drop them somewhere in the repo so clicking on a badge will take you to the report (and further, the report can link out to the problematic decisions/references)

PR #78, opened 2026-08-12
The decision index gains a real Title column

dmarx:
The index table's middle column was one blob — the summary when present, else the title — under a header that said "Title", so any document with a summary showed its summary mislabelled as a title.

Rows now read **code | title | summary | status**:

- The Title column renders `Adr.title` (frontmatter, heading fallback — unchanged logic).
- The Summary column is the `summary:` frontmatter, **honestly empty** when there is none, rather than a fallback quietly repeating the title. An empty cell reads as "this row could say more," which is the truthful state.
- Pipe-escaping (#14) and link rebasing (ADR-005) apply per cell exactly as before; the pipe test now counts five unescaped delimiters, and a new test pins the two-column semantics.

The one-blob shape wasn't accidental — `cell()`'s own docstring said it was kept to make the ADR-004 migration byte-identical. That reason is long since spent; the mislabelled header was what it cost.

390 tests pass, lint clean, `luria index --check` current. The regenerated views are in the diff (collapsed by `.gitattributes`, per #76).

Archaeology on whether this is a regression of something previously fixed is underway; findings will follow as a comment.

---


dmarx:
## Archaeology: not a regression — and the memory has a real source

The session clone was shallow, so I unshallowed to get the full 66-commit history before concluding anything.

**The blob-under-"Title" header is original.** `| # | Title | Status |` arrives in the extraction root commit itself (`216fd5d`, "Luria: the project-memory framework, extracted from strata-g") and is touched by exactly one commit since — this PR. No Summary column ever existed in this repo. And the shape wasn't an accident that slipped in: it's documented as deliberate in three places — `cell()`'s docstring ("a row has always been one blob... what made the migration byte-identical"), ADR-004's summary ("the index is read far more often than the decisions, so `summary` is where the payoff lives"), and the decision template's own frontmatter comment ("the one-blob description the index table shows. Without it the table falls back to the title"). So this PR revises an intentional-but-mislabeled choice whose stated reason (byte-identical extraction from strata-g) is long spent. ADR-004's actual decision — index generated from frontmatter — is untouched; only the column layout moved.

**You're not misremembering fixing titles, though. Two adjacent things really happened:**

1. **ADR-013 / `484f442` (Aug 3): "Name documents for their code; move the title into frontmatter."** A genuine title-fixing sweep — titles became first-class `title:` data, the title-matches-heading lint arrived, 157 links across 32 files were rewritten. Notably, its Context *enumerates* "whatever the index table showed" as one of the three title copies — the table was in view during that fix, but the work addressed where titles live, not what column the index displays them in. Remembering "we already fixed titles" is accurate; the index column just wasn't part of it.

2. **The pending-decisions report has had a real, labeled `Title` column all along** (the #35 committed-reports work). If the memory is of seeing a proper Title column in one of this repo's tables, that's the report table — the decision index never had one until now.

One thing this suggests beyond the fix: the extraction-era "byte-identical" constraint left at least one more fossil in its comments. If you spot other index quirks that feel already-fixed, the same root commit is the likely origin.

---

PR #81, opened 2026-08-15
fix: `legacy-spellings` is nameable in the enforcement dial

dmarx:
Found while porting the migration machinery in #80. One word in a tuple, plus the test that should have caught it.

## The defect

`status_sections()` has emitted a `legacy-spellings` class since rung one landed, but `FAILABLE` never listed it. Both are read by `report_warnings`, which rejects anything in `fail_on` that isn't in `FAILABLE`:

```python
for name in sorted(fail - set(FAILABLE)):
    errors.append(f"luria.toml: `fail_on` names {name!r}, which is no "
                  f"warning class (known: {', '.join(FAILABLE)})")
```

So a project writing `fail_on = ["legacy-spellings"]` was told the class does not exist — while the linter went on printing that very class at it. The class could be reported but never enforced, and the error message actively denied it was real.

That is the enforcement dial refusing a notch it was already reporting on: [DP-1](docs/design-principles.md#dp-1) inside the guard written to catch silent refusals. The guard's logic was right; its vocabulary was incomplete.

## The test

Asserting over the **whole vocabulary** rather than the one class that bit — the same generalization the fix itself wants:

```python
dial_project(project, '"legacy-spellings"')
errors, _ = dial_errors(capsys)
assert not any("is no warning class" in e for e in errors), errors
```

Fired before trusting ([DP-6](docs/design-principles.md#dp-6)): it **fails** against the unpatched `FAILABLE` and passes against the patched one. My first attempt at this test passed on both — it asserted `emitted <= set(FAILABLE)` against a fixture that never produced a legacy spelling, so the subset check was trivially true. Worth noting because that shape looks more thorough than the one that actually works.

Suite: **401 passed**. `luria lint` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_012AWc5urqmvaJUhcvy1VWLM)_

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -80,6 +80,7 @@ The dial between reported and enforced. Status findings are warnings by default;
 | key | default | what it does |
 |---|---|---|
 | `fail_on` | *unset* | Warning classes promoted to failures. |
+| `narrow_terms` | *unset* | This project's own concrete nouns. A title in a scheme marked `titles_generalize` that names one is reported as `narrow-titles`. Luria ships none — the words are yours, and empty means the class never fires. |
 
 ## Schemes — `[luria.schemes.X]`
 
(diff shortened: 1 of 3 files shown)


dmarx:
lgtm

PR #82, opened 2026-08-15
docs: scaffold with `luria new <kind>`, not by copying `_template.md`

dmarx:
Both `_template.md` files and the decisions `README.stub` opened by telling the reader to *copy `_template.md` to the next free number*. That instruction predates `luria new` and had outlived it — `new.py`'s own docstring already says `luria new adr` scaffolds *"the next free decision number, **from `_template.md`**"*. The template was documenting the workflow the command replaced.

It's also worse advice than it looks. `new`'s kinds are **derived from config** — every scheme, fragment directory and journal in `luria.toml` is one, and nothing in `new.py` spells "adr" — so `luria new <kind>` works for a scheme the moment it's declared. And it takes the next free number itself, where hand-copying asks the author to pick one. That's how two branches claim the same number, which is the problem [ADR-049](record/decisions.d/ADR-049.md) then has to solve at the merge point.

## Six files

Fixed in the shipped `template/` scaffold **and** in this project's own record, so an adopter and a maintainer read the same instruction:

- `template/record/decisions.d/_template.md` + `record/decisions.d/_template.md`
- `template/record/principles.d/_template.md` + `record/principles.d/_template.md`
- `template/record/decisions.d/README.stub` + `record/decisions.d/README.stub`

`record/devlog.d/_template.md` already said *"Don't copy this file by hand — run `luria new`"*. It was the only one that had been updated, which is why the divergence was easy to miss.

## How this surfaced

Downstream, in `dmarx/strata-g`. Adding a third document type there meant declaring `[luria.schemes.PC]`, and `luria new pc` worked immediately — at which point it was obvious that its `CLAUDE.md` was telling contributors to hand-copy templates for **all three** of its document types, including a hand-written *"check the highest existing number and increment"*.

strata-g was luria's first consumer, and this is guidance that travelled with the extraction. Fixing it there and not here would leave the canonical scaffold teaching the thing we'd just corrected.

**Verification:** 400 tests pass; `luria lint` clean; regenerated views included. The changes are frontmatter comments and stub prose only — no behaviour — and `publish.yml`'s smoke test already exercises `luria init` → `luria index` → `luria new` against the packaged template.

---
_Generated by [Claude Code](https://claude.ai/code/session_012AWc5urqmvaJUhcvy1VWLM)_

### DIFF
--- a/template/record/principles.d/_template.md
+++ b/template/record/principles.d/_template.md
@@ -1,6 +1,6 @@
 ---
-# Copy this file to DP-<NNN>.md — next number wins. Numbering is sequential and
-# permanent: a principle is cited by number ("per DP-3"), so a number can be
+# Don't copy this file by hand — run `luria new dp`, which takes the next free
+# number. Numbering is sequential and permanent: a principle is cited by number ("per DP-3"), so a number can be
 # retired but never reused. The filename is the code and nothing else; the title
 # goes in `title:` below, where a revision costs an edit rather than a rename
 # plus every link — which matters more here than for a decision, because
--- a/template/record/principles.d/_template.md
+++ b/template/record/principles.d/_template.md
@@ -1,10 +1,15 @@
 ---
-# Don't copy this file by hand — run `luria new dp`, which takes the next free
-# number. Numbering is sequential and permanent: a principle is cited by number ("per DP-3"), so a number can be
-# retired but never reused. The filename is the code and nothing else; the title
-# goes in `title:` below, where a revision costs an edit rather than a rename
-# plus every link — which matters more here than for a decision, because
-# principles are expected to be reworded. Why: LU-ADR-013.
+# Don't copy this file by hand — run `luria new dp`, which assigns the identity
+# and fills in the fields a machine can compute. WHICH identity depends on the
+# scheme's `allocate` mode: `filing` (the default) takes the next free number
+# on the spot, `merge` mints a temporary code that `luria concretize` numbers
+# where merges serialize (ADR-049).
+#
+# Numbering is sequential and permanent: a principle is cited by number ("per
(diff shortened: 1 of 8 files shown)


template/record/principles.d/_template.md
```diff
@@ -1,6 +1,6 @@
 ---
-# Copy this file to DP-<NNN>.md — next number wins. Numbering is sequential and
-# permanent: a principle is cited by number ("per DP-3"), so a number can be
+# Don't copy this file by hand — run `luria new dp`, which takes the next free
+# number. Numbering is sequential and permanent: a principle is cited by number ("per DP-3"), so a number can be
```
dmarx: does it? I thought it gets assigned a temp ID and the number is applied only when the concretization step is run?

dmarx: You're right, and the sentence was wrong in exactly one of the two places I put it.

It depends on the scheme's `allocate` mode, and **the two records this PR touches differ**:

- `template/luria.toml` declares no `allocate`, so both its schemes default to `filing` → next free number on the spot. The claim held there.
- **This** record sets `allocate = "merge"` on both schemes, so the number arrives at concretization. I wrote the `filing` sentence into it anyway.

Verified rather than reasoned — `luria new adr` here returns `record/decisions.d/ADR-tmpohleu.md`.

Rather than write two different comments, both templates and the decisions stub now say `luria new` **assigns the identity** and name both modes:

> WHICH identity depends on the scheme's `allocate` mode: `filing` (the default) takes the next free number on the spot, `merge` mints a temporary code that `luria concretize` numbers where merges serialize (ADR-049).

That stays true whichever mode a project picks, and it teaches that the mode exists — which is what a reader needs *before* a `tmp` code surprises them.

`489766f`.

---

PR #84, opened 2026-08-15
fix(migrate): a moved document lands provisional, not numbered

dmarx:
Found by running the first real two-document promotion downstream, hours after #80 shipped the machinery. **It would have destroyed a document.**

> **Rewritten after review.** The first version added a per-plan counter. That was the wrong level, and this now uses temporary codes instead — see *Why not a counter* below.

## The defect

Every operation in a spec plans against the tree as it is *now* — nothing has moved yet — so `_plan_move` computed `max(target.documents(), default=0) + 1` twice and got the same answer twice. Two `git mv`s onto one path; the second overwrote the first.

The dry-run printed it in plain sight:

```
DP-017 -> PC-001
DP-019 -> PC-001
git mv principles.d/DP-017.md docs/conventions/PC-001.md
git mv principles.d/DP-019.md docs/conventions/PC-001.md
```

## The fix: mint a temporary code, let the concretizer number it

"The next free number" is no more a fact inside a migration than it is on a branch. So `move_doc` now mints a temp code ([ADR-049](record/decisions.d/ADR-049.md)) and `luria concretize` assigns the real one afterwards, where merges serialize — the same bargain `luria new` already makes for a merge-allocated scheme.

The document ends up carrying **both** aliases — the code it migrated from, and the provisional one it wore in between — so citations from either era resolve.

### Why not a counter

A counter is a *second* allocator sitting alongside the one this project already has, with its own arithmetic to get wrong. It also only solves collisions **within** a spec: two branches each running a migration into the same scheme would still collide, and that is precisely the case temporary codes exist for. Reusing the mechanism fixes the general problem and deletes the arithmetic.

## Two things that fell out

- **`Pair.parts`** now returns the new tail as `int` for a rename and `str` for a move, and the sweep skips zero-padding for a temporary tail — which has no padded form to preserve.

- **A pre-existing bug in `luria concretize`.** Its sweep was a case-sensitive `str.replace`, so it upgraded `ADR-tmp47fje` and walked straight past the anchor spelling `#adr-tmp47fje` — leaving a live link pointing at a heading that no longer existed. Generated views are re-derived and were never at risk; a hand-written or migration-written link was. Unreachable until migrations started writing temp anchors into ordinary files, which this change does. The test now asserts `val-tmp` appears nowhere after concretization, not merely that the label updated.

## Also in here

The **0.3.0 version bump**, swept in by an earlier `git add -A` and kept deliberately rather than quietly: `narrow-titles` (#83) and this fix both want a release, and `main` has been sitting at 0.2.0 with unreleased features since #83 merged.

## Verification

The three move tests now assert the full chain — provisional landing, then `concretize`, then numbered — rather than a single numeric outcome. **408 tests pass**; `luria lint` clean.

### DIFF
--- a/luria/migrate.py
+++ b/luria/migrate.py
@@ -73,16 +73,38 @@ class Pair:
     new: str                    # canonical new code, e.g. NEW-004
 
     @property
-    def parts(self) -> tuple[str, int, str, int | str]:
-        """`(old prefix, old number, new prefix, new tail)`.
+    def old_parts(self) -> tuple[str, int]:
+        """Prefix and number. The old side is always numeric: a document that
+        never had a number has nothing to migrate away from."""
+        return aliases_mod.split(self.old)
 
-        The new tail is an `int` for a rename and a temporary-code *string*
-        for a move (ADR-049): a moved document arrives provisional and is
-        numbered by the concretizer. The old side is always numeric — you
-        cannot migrate away from a code that was never assigned."""
-        op, on = aliases_mod.split(self.old)
-        np, raw = self.new.rsplit("-", 1)
-        return op, on, np, int(raw) if raw.isdigit() else raw
+    @property
+    def new_parts(self) -> tuple[str, str]:
+        """Prefix and the LITERAL new tail — `004`, or `tmp47fje`.
+
+        A string either way. The two spellings do not share a type: one is a
+        number carrying a padding convention, the other is an opaque identity
+        (ADR-049). An earlier `int | str` union pushed the ambiguity out to
+        every call site, which then had to test the type to learn which it
+        had — and the padding branch and the provisional branch are not the
+        same question."""
+        prefix, tail = self.new.rsplit("-", 1)
+        return prefix, tail
+
+    @property
+    def new_is_provisional(self) -> bool:
+        """A temporary code, awaiting `luria concretize`."""
+        return not self.new_parts[1].isdigit()
+
+    @property
+    def new_anchor_tail(self) -> str:
+        """How the tail is spelled inside an anchor: `#gp-4`, `#gp-tmp47fje`.
+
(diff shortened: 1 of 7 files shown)


luria/migrate.py
```diff
@@ -73,10 +73,38 @@ class Pair:
...
+        number carrying a padding convention, the other is an opaque identity
+        (ADR-049). An earlier `int | str` union pushed the ambiguity out to
+        every call site, which then had to test the type to learn which it
+        had — and the padding branch and the provisional branch are not the
+        same question."""
+        prefix, tail = self.new.rsplit("-", 1)
+        return prefix, tail
+
+    @property
+    def new_is_provisional(self) -> bool:
```
dmarx, line 95: `is_provisional()` shouldn't live here. this is a test that should be defined in exactly one place and invoked from wherever that is.

dmarx: Agreed, and it was worse than a second definition — it was a *different test*.

`TEMP_TAIL` already existed, composed with a prefix in three places (`Scheme.temp_pattern`, `Scheme.temp_of`, and doc_refs' wikilink matcher), all asking "is this **code** or **filename** temporary?". A caller holding only the tail had nowhere to ask, so `new_is_provisional` invented `not tail.isdigit()` — which answers "this is not a number", and a malformed tail satisfies that too.

`TEMP_TAIL` and `is_temp_tail()` now sit together at module level in `config.py`, with `Scheme.TEMP_TAIL` referencing the constant so every existing regex composition is untouched. `migrate` invokes it.

The new test pins exactly where the cheap version was wrong:

```python
assert not is_temp_tail("nonsense"), "not-a-number is not the same test"
assert not migrate.Pair("DP-004", "VAL-nonsense").new_is_provisional
```

`599679e`. 410 tests pass.

---

Issue #104, opened 2026-08-16
A scheme whose status never varies has a field carrying no information, and nothing says so


dmarx:
Split out of #102, which is about where a status's *meaning* lives. This one is about whether the field is doing anything at all, and it is the half with teeth.

## The observation

Three schemes in one downstream project (`dmarx/mathematics-of-meaning`), each with every record at a single status:

| scheme | records | statuses present |
|---|--:|---|
| claims | 51 | `Active` × 51 |
| arguments | 24 | `Active` × 23, one `Superseded` |
| concepts | 13 | `Active` × 13 |

The first two were bugs, and both were found by a person re-reading rather than by a check.

The claims scheme was the expensive one. Its `fail_on` names `retired-citations`, which makes a retired premise a build failure in every argument resting on it — the enforcement it adopted luria for. Extraction files at `Active` and nothing ever contradicted it, so **nothing was ever retired, so nothing was ever found.** Thirteen green builds across thirteen documents meant only that nobody had judged anything. Twenty-three of those claims had a counterexample exhibited in their own body; two said "It is false as stated" in as many words.

The arguments scheme was the second: its template said status carried a judgment that a `tag_groups` axis was actually carrying, so nine records sat at `Active` tagged `invalid`.

The concepts scheme is still unexamined — 13 of 13 `Active`, with `version:` carrying the movement instead, which is either correct or the third instance.

## Why a check is right here

A status field where every record agrees is indistinguishable from no status field, and the difference matters because other machinery reads it: `active` decides what counts as retired, and `retired-citations` fires off that. A scheme in this state has an enforcement mechanism that cannot fire, and there is currently no signal at all — the build is green precisely *because* nothing is being judged.

It is also cheap to detect and needs no configuration.

## Shape

A status report class, along the lines of:

```
luria: 2 scheme(s) file every record at one status, so `retired-citations` cannot fire there
  CLM: 51/51 Active
  CON: 13/13 Active
```

Notes on getting it right:

- **A report, not an error** (ADR-035). There is no correct proportion, and a corpus whose claims all survive is a legitimate outcome. The finding is "nothing has ever been judged here", which a human should read and may reasonably dismiss.
- **Needs a floor.** A scheme with two records at one status is not evidence of anything; the downstream project's own canary skips below ten.
- Probably worth suppressing for a scheme whose `render = "document"` or whose records are principles — a design-principles scheme where everything is `Active` is the expected state, not a smell.
[...]

PR #108, opened 2026-08-17
lint: give inert-status the acknowledgement every other finding has

dmarx:
## The gap

Every judgment call in luria can be answered where it is raised. A retired citation takes `inactive-ok:`, an unresolvable code takes `unresolved-ok:`, a bare URL takes `url-ok:`, a broad title takes `broad-ok:`. Same bargain each time: the reason is mandatory, it lives at the site, and it renders where the finding would have.

`inert-status` (#104) is the exception, and not by oversight — it is about a *scheme*, and a scheme has no line to comment on. So a project whose uniformity was deliberate had no move available. The check fired on every run, forever, and the only two responses were to ignore it or to make it false.

**A guard nobody can answer is one people learn to skip**, which is the failure mode the acknowledgement directives exist to prevent.

It showed up downstream: `mathematics-of-meaning` has run this warning on every lint invocation for weeks, on two schemes, with no available response.

## The change

`uniform_ok` on a scheme — a mandatory-reason string:

```toml
[luria.schemes.ADR]
dir = "record/decisions.d"
uniform_ok = "Fourteen decisions, none yet superseded — the expected state for a record weeks old, not a dead enforcement field. Revisit at the first supersession."
```

Set, the scheme leaves `inert-status` and appears in a new `acknowledged-uniformity` section, **which still reports the count and the status** and appends the reason. The fact does not stop being true because someone explained it — nothing in that scheme is being judged, and the citation checks still cannot fire off it. It stops being a *finding*.

Three properties, each deliberate:

- **The reason is mandatory.** An empty string is no acknowledgement. Silence carrying no argument is indistinguishable from an oversight, which is what this replaces.
- **The acknowledgement lapses on its own.** It suppresses only while the scheme is *actually* uniform. The first record filed at a second status and the row leaves both sections — no stale directive to clean up, because the condition rather than the annotation is what is tracked.
- **A project cannot promote it to a failure.** `acknowledged-uniformity` is absent from `FAILABLE`, so naming it in `fail_on` hits the existing unknown-class check rather than silently enforcing nothing.

## Rejected (ADR-059)

- **Declaring a one-status vocabulary.** Already suppresses the row by design — "a project saying so on purpose" — but it is the wrong instrument: saying `ADR` has one status asserts that a decision can never be superseded, which is false and would hide the finding the day it became real. Suppression and claim have to be separable.
- **A boolean flag.** Works mechanically, drops the only part worth keeping.
- **Dropping the check.** It is right about what it reports. Downstream, of the two schemes it fires on, one is a young record and the other turns out to hold a concept that project's own claims say should be retired.
- **Raising `FLOOR`.** Trades one arbitrary threshold for another and still leaves a project no way to speak.

## Verification

462 existing tests pass. Five new in `tests/test_statuses.py` pin: the check still fires without the acknowledgement; the row moves from finding to note *and keeps the count, the status and the reason*; the acknowledgement lapses when a second status appears; and the class is not promotable to a failure.

`docs/configuration.md` regenerates from the schema and picks the key up automatically.

## Downstream

`dmarx/mathematics-of-meaning` uses it on both its `ADR` and `CON` schemes. The `CON` reason is worth reading as a worked example of what the field is for — it records an audit of all 18 concepts, explains why status does little in a scheme whose records adjudicate vocabulary rather than assert propositions (six are already at `version: 2`, none has ever changed status), and names the live case that *would* retire one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DpK3k2WhH9Pnc3i6TjLMVv)_


dmarx:
Rebased onto main and reshaped per review.

**The ADR is folded into ADR-057 as version 2 rather than filed separately.** Widening a check to include the reply to it is the same choice restated, not a new decision — so ADR-057 goes to `version: 2` with a `history:` entry saying what changed, and adopts as `Active`. Version 1 shipped a guard nobody could answer, which is the failure mode the acknowledgement directives exist to prevent; adopting the check and its reply together is the honest shape.

That also resolved a collision worth naming: **both sides had filed an ADR-059.** Main's is the configuration-reference decision; this branch's was the acknowledgement. An add/add conflict on a sequential code is exactly what merge-time allocation (ADR-049) exists to prevent, and this branch predates it. Folding rather than renumbering means main's ADR-059 keeps its number and nothing that cites it moves.

Other conflicts: `config.py` hand-merged so `uniform_ok` sits alongside the `tags_file` and `references` fields from #121, threaded through the scheme builder main introduced; generated views regenerated; and the changelog fragment this branch modified had already been collected on main, so the feature gets a fresh one.

508 tests pass. `ADR-056` and `ADR-061` still read `Proposed` in this branch's lint output because it predates the status sweep in #126 — both land `Active` when that merges.

PR #120, opened 2026-08-25
feat: warn when a commit message suppresses its own build

dmarx:
Independent of #119 — based on `main`, no overlap.

## The convention has no escape sequence

There is no way to mention a CI skip marker in a commit message without invoking it. So a message *explaining* what the marker does tells the forge not to build the commit that explains it.

That happened this week to a project adopting the recommended workflow — in the commit that adopted it, written by an author who had read the caution in the scaffolded `docs.yml` header **first**. A warning that is read and then tripped anyway is weak evidence for documentation as the remedy.

## Why it is expensive out of proportion to its silliness

A suppressed run is not a red build. It is *no* build, and the pull request goes on displaying the previous commit's green checks — because those are the most recent checks that exist. Silence is indistinguishable from success unless somebody thinks to ask which commit the green belongs to, and nobody asks that on a PR that looks passing.

In the reported case it surfaced only because a polling script kept returning stale SHAs.

## `luria skip-markers`

Reports commits carrying a marker in the message **body**.

**Position is the entire rule**, and it is what keeps this quiet enough to be worth running. Someone instructing the forge puts the marker where the convention puts it — subject line, or a trailer — and so does every tool that generates one, including this package's own `generate` action. Someone writing *about* it lands the marker mid-paragraph.

So `docs: regenerate views [skip ci]` never fires: a one-line message is first and last at once. No author list, no bot identities to maintain, no configuration.

A note on what the position claim actually is: GitHub documents the marker as honoured in the first or last line. Measured, it is broader — the reported case had it on line nine of a body and every workflow was still skipped. The rule here is therefore not a claim about what the forge honours; it is a claim about **what the author meant**, which is the only thing worth guessing at.

## Kept quiet on purpose

`ci.py` already carried the lesson this needed, in a comment about a helper built and removed before merge: *a warning that fires on correct runs trains readers to skip warnings.* So this:

- prints nothing when there is nothing to say
- warns rather than fails (`--strict` exists, and is off)
- stays silent when it cannot read history — `actions/checkout` fetches depth 1 by default, and a build must not break over a checkout setting

That last one is itself a small trap: a guard that sees nothing looks exactly like a guard that found nothing. Which is why the scaffolded workflow now sets `fetch-depth: 25` on the lint job rather than leaving it to be discovered.

## What it cannot do

**It cannot catch the commit it is about.** The suppressed run is the one that did not happen, so nothing inside it can speak — the warning necessarily arrives on a *later* run, once history has moved past the offending commit. This is a backstop, not a fix, and the docs say so rather than implying a guarantee.

The honest fix is client-side: a `commit-msg` hook that refuses a marker in a body. That is a bigger conversation, because scaffolding git hooks into someone's repository is intrusive in a way writing files is not. Not doing it here; noted in the devlog as the better shape if you want it.

## Changes

- `luria/ci.py` — `SKIP_MARKERS`, `prose_skip_marker()`, `commits()`, `run()`
- `luria/cli.py` — registered, and listed under the CI-only commands
- `actions/lint/action.yml` — new `skip-marker-range` input; empty string turns it off
- `template/.github/workflows/docs.yml` — `fetch-depth: 25` on the lint checkout, and the header caution now says a check exists and what it cannot do
- `docs/cli.md` — the command
- `tests/test_skip_markers.py` — 17 tests

498 pass, `luria lint` clean. Filed with a changelog fragment and a devlog entry.

### DIFF
--- a/actions/lint/action.yml
+++ b/actions/lint/action.yml
@@ -25,6 +25,14 @@ inputs:
   reports-path:
     description: Where `luria reports` writes — your luria.toml's `reports` path.
     default: docs/reports
+  skip-marker-range:
+    description: >-
+      Commit range scanned for skip markers written as prose — a message
+      *describing* the convention carries the marker and so suppresses its own
+      build. Empty string turns the check off. It needs history to read:
+      `actions/checkout` fetches depth 1 by default, so give the checkout a
+      `fetch-depth` that covers this range or the check has nothing to see.
+    default: HEAD~20..HEAD
 runs:
   using: composite
   steps:
@@ -40,6 +48,13 @@ runs:
     - if: ${{ always() }}
       shell: bash
       run: luria reports
+    # Independent of the lint's verdict, so `always()`: a suppressed build is
+    # worth reporting on the run that failed as much as on the one that
+    # passed. Warns and does not fail — the marker is legitimate on the
+    # commits that mean it, and only position distinguishes them.
+    - if: ${{ always() && inputs.skip-marker-range != '' }}
+      shell: bash
+      run: luria skip-markers --rev-range "${{ inputs.skip-marker-range }}"
     - if: ${{ always() && inputs.reports-artifact != '' }}
       uses: actions/upload-artifact@v4
       with:
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -116,6 +116,32 @@ Assemble fragment directories into their documents — a changelog from
 Fragments exist so the assembled file is not a lock every branch must touch.
 Collect on a cadence, not on every merge.
 
+## `luria skip-markers [--rev-range R] [--strict]`
+
+Report commits whose message carries a CI skip marker **in its body** — a
+message *describing* the convention contains the marker, and so tells the forge
+not to build the commit that describes it.
+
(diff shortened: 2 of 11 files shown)


dmarx:
Closing: measured, and it comes out backwards. The reasoning is recorded as an ADR in #125 so it outlives this thread.

**It cannot fire in the case that does harm.** When the marker sits on a commit that is the tip and stays the tip, no workflow runs — so nothing exists to run the check inside. Both incidents that motivated this PR were that shape, and the checker would have been silent for both.

**It fires when nothing is wrong.** A probe pushed two commits together, the first mentioning the marker in its body and the second clean at the head. The runs fired normally: GitHub reads the head commit only, so a marker anywhere else is inert. The checker would have reported that commit anyway, and its message had to hedge about whether the commit was ever a tip — a warning admitting it does not know. `ci.py` already carries the note about why that alone is disqualifying.

Required status checks in branch protection are the answer, and are not something this package can ship: everything here runs inside a workflow, and the failure is the workflow not running.

The half worth keeping already landed with #119 — `adopting.md` documents the hazard and the detail that makes it findable, that the failure presents as no build rather than a red one.

PR #124, opened 2026-08-25
probe: does a non-head skip marker suppress a PR run?

dmarx:
Throwaway. Two commits: the first mentions the skip marker in its body, the second (the head) is clean. Testing whether the marker on a non-head commit suppresses the pull_request run, which decides whether #120's warning has false positives. Will be closed immediately.


dmarx:
Answered: runs fired on the clean head commit despite the non-head commit carrying the marker in its body. GitHub checks only the head. Deleting.

PR #129, opened 2026-08-25
feat: luria init takes shorthand for extra schemes and journals

dmarx:
A project that wanted the defaults **plus one scheme** had two options: take the shipped template exactly, or write a whole `luria.toml` and pass `--config`. There is a large middle, and reaching it meant writing four lines of which three follow from the prefix.

```console
$ luria init --issue-url https://github.com/you/proj/issues \\
             --schemes "RFC,SPEC:document" --journals "incidents:day"
```

Each entry is `NAME` or `NAME:kind` — `index`/`document` for a scheme, `year`/`month`/`day` for a journal. Paths follow the prefix: `RFC` → `record/rfcs.d` → `docs/rfcs`.

## The design decision

**The shorthand is an argument, never a stored format.** What lands in `luria.toml` is the ordinary commented table:

```toml
# RFC — browsed one at a time, so its view is an index plus tag pages.
# The paths follow the prefix; rename them if this family is better
# called something other than what its codes spell.
[luria.schemes.RFC]
dir    = "record/rfcs.d"
output = "docs/rfcs"
render = "index"
```

No second grammar, no precedence question between a compact form and an explicit one, nothing new for `config.py` to parse. The saving is in what somebody types once — and the file a contributor or an agent later reads looks like every other project's.

**Additive, and that is load-bearing.** A declared family replaces the shipped one whole (ADR-047), so *"the defaults plus an RFC scheme"* works only because the template's own ADR and DP tables stay in the file. That puts the rule in front of the reader at the moment it matters. Removing a default is deleting a table.

**Refused** where a config already exists, and alongside `--config`: the shape is declared there, and a flag should not append to a file the project owns.

## Verified end to end

On a fresh repository — one command, **23 files, four schemes, two journals**, `luria index` and `luria lint` clean, and `luria new rfc` scaffolding `RFC-001` immediately. The error paths too: an unknown render names the vocabulary, redeclaring `ADR` is refused, and so is shorthand against an existing config.

## A bug the tests caught

The helper took its default as the first of the listed kinds, so a journal defaulted to `year` while luria's own default is `month` — a display order tied to a fact about the dataclass. The default is passed explicitly now, and the docstring says why.

## Not included

Removing a shipped family, and the collocated layout. Both change the template rather than extend it, and both are one edit to a file the user can now see. `--config` and `examples/` remain the right tools for a whole shape.

## Checks

522 tests pass (14 new); `luria lint` clean. ADR filed with a temporary code.


dmarx:
Folded into #130, which contains these commits — I branched the follow-up work without returning to main, so #130 already carries the shorthand and the issue-URL inference. Reviewing them as one change is the right call: they are the same story, and #130 adds `luria config` and the render docs on top.

PR #131, opened 2026-08-25
test: acknowledge the ADR-007 fixture, clearing the reference-status report

dmarx:
The report had exactly one finding, and it was fixture data rather than a citation.

`tests/test_prose_frontmatter.py` writes an `ADR-007.md` into a temporary project and checks that a bare reference to it gets linkified. The number collides with this repository's own `ADR-007`, which is **Superseded** — so five lines of fixture data read as citations of a retired decision.

Acknowledged at file level, saying what they actually are.

**The report now reads clean in every section:**

| | |
|---|---|
| cited while not in force | 0 unacknowledged (41 vouched for) |
| codes resolving to nothing | 0 (56 marked deliberate) |
| stale annotations | none |

## What this is an instance of

Worth recording rather than treating as a one-off. Test fixtures across the suite are drawn from the live ADR sequence — **37 distinct numbers over 15 files** — and I checked where they stand:

- 4 real ADRs are not in force today; only `ADR-007` was borrowed, which is why there was exactly one finding.
- **35 fixture codes name decisions that are currently `Active`.** Each becomes a report finding on the day its decision is retired. That is precisely how this one arrived: `ADR-007` was fine until it was superseded.

ADR-034 named the shape — *"any specimen drawn from the sequence is a collision waiting for the sequence to arrive"* — and #123 fixed it for the principles schemes, where a rename would have swept the fixtures too.

The decision fixtures are the larger half, and I left them alone. Converting them means fifteen files with assertions on `ADR-` strings throughout, which is its own piece of work rather than a line in a report cleanup. Worth an issue if you want it done; the cost of not doing it is one acknowledgement each time a decision retires.

## Checks

541 tests pass; `luria lint` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ generated and linted by the CLI it ships.
 
 <!-- luria:badges -->
 [![needs decision: 0](https://img.shields.io/badge/needs%20decision-0-brightgreen)](docs/reports/pending-decisions.md)
-[![cited, not in force: 1](https://img.shields.io/badge/cited,%20not%20in%20force-1-orange)](docs/reports/reference-status.md)
+[![cited, not in force: 0](https://img.shields.io/badge/cited,%20not%20in%20force-0-brightgreen)](docs/reports/reference-status.md)
 <!-- /luria:badges -->
 
 ## Kinds of record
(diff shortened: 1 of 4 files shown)


dmarx:
Follow-up filed as #132 for the pattern behind this fix.

One correction to the description above: it says fifteen files. That count came from matching filename-shaped occurrences (`ADR-0NN.md`) only — the true figure is **33 files**, 37 distinct codes, 35 of which name currently-`Active` decisions. The other numbers here are unaffected, and #132 carries the measured ones.

PR #134, opened 2026-08-25
fix: name UTF-8 at every file read and write

### LINKED ISSUE #112: Every file read and write in the package is locale-dependent, so `luria index` crashes on Windows out of the box
## What happened

On Windows with a default (non-UTF-8) locale, `luria index` crashes on a repo that `luria init` created thirty seconds earlier:

```console
$ luria init
wrote 18 file(s), skipped 0 existing.
$ luria new adr --title "Cache invalidation is time-based"
$ luria link --fix
$ luria index
  File "...\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input, self.errors, encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '✅' in position 3044:
  character maps to <undefined>
```

The character is U+2705 WHITE HEAVY CHECK MARK, headed for a generated view. `cp1252` cannot represent it, so the write dies.

## The fix that isn't

Setting `PYTHONUTF8=1` on the *same* tree fails the other way:

```console
$ PYTHONUTF8=1 luria index
  File "<frozen codecs>", line 325, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x97 in position 45: invalid start byte
```

Byte `0x97` is an em dash — written as cp1252 by luria's own earlier run, and now unreadable to luria under UTF-8. The two halves of the same bug disagree with each other.

`PYTHONUTF8=1` from a **clean** repo does work end to end. That is the current workaround, and it has to be set before the first command, not after.

## Root cause

There is no explicit encoding anywhere in the package. From `luria` 0.3.0 as installed from PyPI:

```console
$ grep -rno 'read_text()\|write_text(' *.py | wc -l
82
$ grep -rn 'encoding=' *.py | wc -l
0        # the two urllib calls in remotes.py are urlopen, not text I/O
```

`Path.read_text()` / `Path.write_text()` / `open()` with no `encoding=` use `locale.getpreferredencoding()`, which is `cp1252` on a default Windows install. Every read and every write in luria is therefore locale-dependent, and the record is full of em dashes, arrows and check marks.

Python 3.15 will make UTF-8 mode the default ([PEP 686](https://peps.python.org/pep-0686/)), which fixes this eventually. Until then, and for 3.11–3.14 which luria supports today, Windows is broken out of the box.

## Suggested fix

Pass `encoding="utf-8"` at all 82 sites, or route text I/O through two helpers so there is one place to get it right ([DP-4](https://github.com/dmarx/luria/blob/main/docs/design-principles.md#dp-4)). Newline handling is worth deciding at the same time: `write_text` on Windows translates `\n` to `\r\n` unless `newline=""` is passed, which will show up as diff churn on any repo edited from both platforms.

A guard is available cheaply. Every job in `ci.yml` is `runs-on: ubuntu-latest` at `python-version: "3.12"`, so nothing in the suite can observe this — the encoding is right on the runner by accident of platform, and a green build says only that nobody has looked ([DP-6](https://github.com/dmarx/luria/blob/main/docs/design-principles.md#dp-6)). One `windows-latest` job, or a Linux job run under a non-UTF-8 `LC_ALL` with `PYTHONUTF8=0`, would fire it.

## Environment

- luria 0.3.0 from PyPI, clean venv
- Python 3.14.4, Windows 11, cp1252 locale
- Reproduced twice from scratch, in `%TEMP%` repos with nothing but `git init`

dmarx:
Closes #112.

## Reproduced without Windows

`LC_ALL=C` gives an ASCII preferred encoding and the same failure the report describes, which is what made this testable in CI:

```console
$ LC_ALL=C PYTHONUTF8=0 luria index
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 100
```

The root cause is as reported: **100 file I/O calls in the package, 7 of which named an encoding.**

## Two halves, opposite treatments

**Files are UTF-8, unconditionally, at every call site.** A record is plain files other people clone; one that only opens on the machine that wrote it is not serving its purpose. There is no case where the platform's guess is right for a file.

**The console keeps the platform's encoding and only stops raising.** `luria init` prints an arrow; under a cp1252 or ASCII terminal that used to be a traceback instead of a scaffold. Both streams now reconfigure with `errors="replace"`, so the character becomes `?` and the line still arrives.

The asymmetry is the decision, and it is recorded as one. Writing UTF-8 at a cp1252 console would trade a crash for mojibake — a user who cannot read the output is no better off than one who got a traceback.

## The guard earned its place immediately

The structural test walks the package's syntax tree and fails on any `read_text`/`write_text` without an explicit encoding.

My first mechanical pass used a regex and left **five nested calls** untouched — in `view.write_text(collect(view.read_text(encoding="utf-8"), …))` the inner call's `encoding=` made the outer one look already covered. The AST test named all five, and an AST-guided second pass fixed them.

## Verified end to end

Under `LC_ALL=C PYTHONUTF8=0`, from an empty directory: `init`, `new adr`, `link --fix`, `index`, `lint`, `reports` and `site` all succeed, and the generated report is UTF-8 containing the `✅` that used to kill the write. Two of the three new tests run that in a subprocess, because the interpreter settles its encodings at startup and no fixture can undo that.

## What this does not fix

A tree written by an older version under cp1252 still holds bytes this version refuses to read. Nothing here repairs one, and the failure is at least loud. The population is a Windows user on a pre-fix release, and the remedy is `git checkout` of the generated views, which are derived anyway. Recorded in the ADR's consequences.

## Checks

558 tests pass (3 new); `luria lint` clean. ADR filed with a temporary code.


dmarx:
Added a design principle, since the encoding fix turned out to be one instance of a pattern this project has been following without stating it.

> **A record must not constrain the project it records**

A record holds claims *about* a project, which makes the project the variable — so the recorder has no standing to narrow it: not its language, its operating system, its forge, or the format the record is stored in.

Four places it has already been tested, all of them in the tree rather than aspirational:

| | |
|---|---|
| **Language** | source files are scanned as text, with no parser and no list of languages — adding a language means adding a glob |
| **OS / terminal** | files UTF-8 unconditionally; the console left at the platform's encoding and merely stopped from raising |
| **Forge** | an issue URL inferred for hosts whose issue paths are known, and not at all for the rest |
| **Storage** | markdown is what a record looks like today and appears nowhere in the model |

The discipline underneath is a pair, and both halves are load-bearing:

> Be explicit in what you write, and forgiving in what you assume.

A file has a reader on another machine and has to open there. A console has exactly one reader, and guessing wrong should cost a `?` rather than the command.

**The failure mode is the part worth writing down.** Nobody decided this package required a UTF-8-capable console. It never said what it needed, and a hundred call sites inherited whatever the platform preferred — fine on the machine where the code was written, a stack trace on somebody's first run. So the question is not *did we choose to support this environment*, it is *did we assume anything we never wrote down*. The first gets asked during design; the second only by a stranger's bug report.

The corollary is the expensive half, and it is stated rather than glossed: no per-language parsers, no format-specific model, no guessing at an unknown forge. Each is real precision given up, bought back in reach.

Filed `Active`, citing the encoding decision it came from — both carry temporary codes and concretize together.

dmarx:
Reframed the principle per review — it was a prohibition wearing a principle's clothes.

> **Meet the project where it is**

A project picks its language, its platform, its forge and its file layout for reasons that have nothing to do with keeping a record, and it made those choices before this tool showed up. The record is the guest.

The negative framing also failed a case the positive one handles: a self-hosted forge gets a record that works and one field to fill in. Under "must not constrain" that reads as a violation; under "meet the project where it is" it is the aspiration partly met, which is what it actually is.

Evidence, discipline and cost are unchanged — the four tested places, *explicit in what you write and forgiving in what you assume*, and the bill paid in precision given up for reach. The failure mode is reframed to match: not a rule being broken, but an aspiration quietly slipping, because coupling rarely arrives as a decision.

PR #150, opened 2026-09-03
ADR: a workflow file cites a decision by number or by prose, never by a temporary code

dmarx:
The decision behind #149's guard, filed as `ADR-tmpvs0gt`, Active — the mechanism is already merged, and this records why it took that shape.

## What it says

`luria concretize` rewrites a temporary code everywhere the code globs reach, workflow files included (on purpose: their comments cite decisions and the reference lint should see a retired one there). GitHub refuses a push from the workflow token that modifies `.github/workflows/`. A temporary code is the one citation the job must later rewrite, in the one place it can never write — so the first merge under ADR-068 had its generation commit rejected whole, and the default branch went red with nothing wrong in the record.

**Decision:** a workflow file cites a decision by number or by prose; a temporary code there is a lint error naming that remedy.

**Rejected:** a concretizer that skips workflow files (leaves a dangling code the record can never resolve, and `--check` needs the same exception); dropping workflow files from the code globs (loses the retired-citation check on the comments that most need it); a token with `workflows` permission (a credential the record's correctness would then depend on, and a fresh scaffold would fail until someone created it).

**Named limit:** a *numbered* citation in a workflow file that a migration renames hits the same refusal; that is the migration PR's to edit by hand. This decision covers the temporary code, which is the case that recurs.

## Verification

- `luria index` then `luria lint` in the working tree: clean. The branch carries no regenerated views.

https://claude.ai/code/session_01NCmBUY14BKhLGbHNKLuP6r

---
_Generated by [Claude Code](https://claude.ai/code/session_01NCmBUY14BKhLGbHNKLuP6r)_

### DIFF
--- a/record/decisions.d/ADR-tmpvs0gt.md
+++ b/record/decisions.d/ADR-tmpvs0gt.md
@@ -1,0 +1,140 @@
+---
+# Don't copy this file by hand — run `luria new adr`, which assigns the
+# identity and fills in the fields a machine can compute. WHICH identity
+# depends on the scheme's `allocate` mode: `filing` (the default) takes the
+# next free number on the spot, `merge` mints a temporary code that
+# `luria concretize` numbers where merges serialize (ADR-049). The kinds are the
+# config: every scheme, fragment directory and journal in luria.toml is one, so
+# `luria new <kind>` works for a scheme the moment it is declared.
+#
+# Numbering is sequential and carries information (it's the order decisions were
+# made). The filename is the code and nothing else; the title goes in `title:`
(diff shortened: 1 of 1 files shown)


dmarx:
1. "Never" is incorrect and makes it so even the user can't run concretize on the workflow file. The issue is just the bot's ability to modify the workflow (do we just need to add a permission to a token to make that possible?). Make it about that. 
2. As I mentioned earlier today: you should be favoring positive affirmative language in decisions and only reserving negative constraining language (e.g. "never") for cases where you are absolutely certain your use is correct. fuck it: turn this guidance into a decision too so we can reference it.

Issue #171, opened 2026-09-05
Typed edges exist but a sequence does not: render a transitively-walked relation as a chain view


dmarx:
## Where this stands

`edges.py` already does most of the work. A reference field is a typed
relation, `superseded_by` and `influenced_by` are built in, `Graph` gives
`outbound`/`inbound` by code, and the site renders both directions on every
page. Declaring

```toml
[luria.schemes.LIT.references]
extends          = { scheme = "LIT", required = false, many = true }
compared_against = { scheme = "LIT", required = false, many = true }
```

works *today*: the codes are checked, resolved, and become edges.

What does not exist is any view that walks such a relation **transitively**.
`inbound`/`outbound` answer "what is next to this document"; nothing answers
"what is the sequence this document is a step in", which is the thing worth
reading.

## Why the neighbour view is not enough

The anthology has roughly ten named sequences and 65 of its ~330 documents
use chain language. Two examples, at the length a reader actually wants:

- plain residual → Hyper-Connections → mHC, with Attention Residuals and
  Gated Residual as siblings and oHC as a critique
- warm restarts → single cosine cycle → warmup-stable-decay →
  decay-to-zero

Rendered as neighbours, that is four pages each saying "extends: X" and
"extended by: Y", and the reader reassembles the line by clicking. Rendered
as a chain it is one ordered list, on one page, with the siblings branching
off it — which is the form the prose in those documents is already
laboriously imitating, once per document, by hand.

And the prose form has already failed in the way duplicated prose fails:
LIT-134 and LIT-140 each ended with a sentence saying the two residual
replacements had never been compared against each other. True when written,
false three weeks later, corrected in two places, found only because someone
read both notes the same afternoon. `compared_against:` on the *third*
document makes that claim a derived, always-current absence instead of a
sentence two people have to remember to delete.

## Proposal

```toml
[luria.chains.residual]
scheme   = "LIT"
relation = "extends"
sibling  = "compared_against"     # optional: cross-links rendered as branches
output   = "docs/chains"
```

`luria index` walks `relation` transitively over the scheme's graph, groups
documents into maximal weakly-connected sequences, and renders one page per
chain: the spine in order, each step's title, status and one-line summary,
with `sibling` edges shown as branches off the step they attach to. A
document in no chain renders nowhere; a chain of one is not a chain.

Two structural findings fall out of the walk and belong in the lint, because
each is always wrong and mechanically detectable:

- **a cycle** — A extends B extends A. Not a judgement call.
[...]

Issue #169, opened 2026-09-05
A scheme's `_template.md` is never checked against the scheme's own contract, so the scaffold can quietly contradict the schema


dmarx:
## What happened

The anthology decided (ADR-010, following #141) that a practice may name
several sources: one recommendation frequently rests on a paper, its
replication, and the production model that shipped it, and recording only the
first was throwing away the strongest part of the evidence. The config was
changed accordingly:

```toml
[luria.schemes.SOTA.references]
source = { scheme = "LIT", required = true, many = true }
```

The lint honours it. `many` resolves every element and makes each an edge.
Everything upstream works.

Three weeks and 34 new documents later, here is the distribution of sources
per practice across 144 practices:

| sources | practices |
|---|---|
| 1 | 140 |
| 2 | 1 |
| 3 | 1 |
| 7 | 1 |
| 8 | 1 |

The four are the ones a person hand-edited while writing the ADR. Every
practice filed *since* the decision is single-sourced — not because the
authors disagreed with it, but because `record/practices.d/_template.md`
still says:

```yaml
# REQUIRED. The reading note for the paper this comes from — `LIT-000`.
source: LIT-000
```

A scalar. The contract says list; the form says one; the form is what people
fill in. The capability is live and inert.

## Why this is luria's problem and not just a stale template

`_template.md` is excluded from every pass in the codebase, deliberately and
by name — `Scheme.is_template` calls it "a form, not a document",
`ref_status` skips it, `site.ALWAYS_EXCLUDED` skips it, `lint` skips it. That
is right for the *document* checks: a template's `LIT-000` should not be a
broken reference and its `SOTA-NNN` should not be an unresolved code.

But it means the template is the only file in the record that states the
schema and is never compared to it. And it is the highest-leverage file in
the record, because every document is a copy of it. A drift between
`luria.toml` and `_template.md` doesn't produce one wrong document; it
produces every subsequent document, in the wrong shape, silently, and the
lint reports each of them as fine — because each one *is* fine. A scalar
`source:` on a `many` field is legal input.

This is the same failure `narrow-titles` and `inert-status` were added for:
a declaration that nothing exercises. The difference is that here the
declaration is exercised — by four documents out of 144 — which is exactly
enough to look adopted.

## Proposal: a `template-drift` finding

Compile the scheme's contract as usual, then check the template's
frontmatter against its *shape*, not its content:

- a field the contract declares `many` must be scaffolded as a YAML list
- a field the contract declares required must be present in the template
[...]

Issue #170, opened 2026-09-05
A record can say what it believes but not what would change its mind: conditional field requirements (`required_when`)


dmarx:
## The gap

Every scheme has a status vocabulary and a lint that checks it. Nothing
anywhere records **what would move a document from one status to another**.

That is not a hypothetical shortfall. Here is what it cost the anthology this
week.

SOTA-133 and SOTA-136 are two competing replacements for the residual
connection. They were filed a day apart, both `Proposed`, with conditions
stated in prose that are almost word-for-word identical:

> Why *Proposed*: one group's paper and that group's production model.
> Promote on an independent result.

An independent result arrived — a third party's design study that ran both
against its own architecture. By the letter of both conditions, both
practices promote. Only one did, because on reading the paper it turned out
to *test* one of them as a design and merely *mention* the other as
comparable at scale. The condition counted papers when what mattered was what
the papers were about.

Two things went wrong, and they are different problems:

1. The condition was in prose, so nothing pointed at it when the evidence
   arrived. It was found because a person happened to be reading both
   documents that afternoon — the same accident that
   [ADR-011 in the anthology](https://github.com/dmarx/anthology-of-the-sota/blob/main/record/decisions.d/ADR-011.md)
   documents for lineage prose.
2. The condition was phrased as a *quantity* of evidence rather than a *kind*
   of evidence, and a quantity is satisfiable by evidence pointing the wrong
   way.

Only (1) is luria's to fix. But (1) is what makes (2) invisible: a condition
nobody rereads is never revised.

Across 144 practices, **four** state a promotion condition at all. Eight are
`Proposed`, one `Deferred`. So most non-active documents in that record say
they are provisional and say nothing about what provisional is waiting for —
and there is no way for a reader, or a lint, or a person arriving in six
months, to tell a document that is waiting on a specific result from one that
was filed hesitantly and forgotten. Both render identically.

## Proposal

A field can already be required (`requires`), required-as-a-group
(`field_groups`), typed (`references`), or drawn from a vocabulary
(`fields.NAME.vocabulary`). Add: required *conditionally on another field's
value*.

```toml
[luria.schemes.SOTA.fields.promote_when]
required_when = { status = ["Proposed", "Deferred"] }
```

Read as: a practice at either of those statuses must carry `promote_when:`.
An `Active` or `Superseded` one need not — nothing is waiting.

This composes with what exists rather than adding a parallel mechanism.
`contract.py` already exists precisely so that "what does this scheme demand
[...]

Issue #141, opened 2026-09-03
Proposal: Semantic Contracts and Composable Facets for Luria


dmarx:
## Summary

Extend Luria by allowing the semantic values already present in a record—scheme, status, tags, and eventually other controlled vocabularies—to contribute **small structural constraints**. Luria composes those constraints into an **effective contract for each entry**, retaining provenance for every obligation.

This contract becomes a shared semantic substrate for validation, explanation, construction, migration, and eventually typed graph operations.

The proposal deliberately avoids introducing a general-purpose rule language or exposing dependent type theory directly. Instead, it strengthens abstractions Luria already has:

* schemes define record classes;
* `statuses.yaml` and `tags.yaml` define scheme-local semantic vocabularies;
* `luria.toml` defines system wiring and combination rules;
* `_template.md` provides human-oriented construction scaffolding;
* references already give frontmatter fields stronger semantics than mere presence checks.

Luria already establishes useful precedents for this architecture. Status meanings are scheme-relative and deliberately live in YAML beside the records, while rules about combining vocabulary values belong in `luria.toml`. Tag-group membership can already be derived from `primary_for` metadata in `tags.yaml`, explicitly to avoid duplicating semantic facts across configuration surfaces.

The proposed direction is therefore evolutionary rather than architectural replacement.

---

# Design Principle

The central semantic operation should be:

$$
\Gamma,e \;\longrightarrow\; C(e)
$$

where:

* \(\Gamma\) is the distributed semantic environment of the project;
* \(e\) is an entry;
* \(C(e)\) is the effective contract induced by the entry's scheme and selected semantic values.

The contract is checked against the entry:

$$
e \models C(e)
$$

but can also be inspected, diffed, and used to construct or repair entries.

The intended user-facing mental model is simpler:

> **Selecting a meaningful status or tag may bring obligations with it. Luria can tell you what those obligations are and why they apply.**

---

# Proposal 1 — Introduce an Internal Effective Contract

Add an internal representation for the structural obligations applying to one entry.

This should initially be an implementation abstraction, not a new user-facing concept.

A minimal normal form might contain:

```text
Contract
  fields
    <name>
      presence: optional | required | forbidden
      type: Any | Ref[SCHEME] | ...

  sections
    <heading>
      presence: required | forbidden

  cardinalities
    ...
```

Every atomic constraint must retain provenance.

For example:

```text
field: evidence
presence: required
type: Ref[LIT]

because:
  tag = empirical

declared at:
[...]

PR #172, opened 2026-09-06
Template drift, chains, and chain annotation

dmarx:
> [!NOTE]
> **Split done** — `required_when` is now [#177](https://github.com/dmarx/luria/pull/177), on `main`, and this PR is rebased onto it. Base is `claude/required-when-e6evzh`; retarget to `main` once #177 merges and the diff here shrinks to what it says on the tin. The rebuilt tree is byte-identical to the pre-split one, and the same 845 tests pass.

Three issues, each measured on a real corpus rather than imagined.

## `template-drift` (#169)

`_template.md` is exempt from every document check, correctly — and that left it the only file in a record which *states the schema* and is never compared to it. It is also the file every document is a copy of, so a drift between `luria.toml` and the form does not produce one wrong document. It produces every document filed afterwards, and the lint calls each of them clean, because each of them is: a scalar in a `many` field is legal input.

The consumer record measured it. Three weeks after deciding its `source:` field was plural, **140 of 144** entries were single-sourced; the four exceptions were the ones a person hand-edited while writing the decision. Fired on that record, the check produces exactly one row, on the one file that is wrong.

Two holes in `luria new` found while wiring it: a comma-separated value rendered as `source: 'LIT-1, LIT-2'` — the scaffolder writing the one shape the lint rejects — and a field the form did not already scaffold was **silently dropped**, with the command printing a path and exiting 0.

## `[luria.chains]` (#171)

Typed edges gave every page its neighbours; nothing answered *what sequence is this a step in*. So the sequences lived as prose, re-described once per participant, and two notes went stale on the same fact — each claiming two rival designs had never been compared — corrected in two places by luck.

The page renders order, title and status and nothing else: **the field carries the sequence, the prose keeps the argument.** `broken-chains` reports a succession that loops and a comparison only one side declares; neither is acknowledgeable.

Three things the implementation had to get right: a line is a DAG rather than a list, a rival that succeeds nothing still joins through its cross-link, and both graph walks terminate on a cycle — a finding is not a licence to render nothing.

Four bugs the first real corpus found that 20 green tests did not: every rendered link was broken (it pointed into the scheme's *view* directory, which holds a README and tag pages and never a page per document), a status note's link went out unrebased, the page had to register in `is_generated`, and `extends:` counted as citing a retired document — once per retired step in every chain, forever, at the field whose job is to name it.

## `annotate` (#173)

A chain rendered an agreed trunk and a disputed fork identically. `annotate = "consensus"` shows a second declared field beside the status, read through the contract so a `default` shows its default rather than a blank.

```
- SOTA-137 — freely learned mixing              (Superseded, unassessed)
  - SOTA-136 — doubly-stochastic mixing         (Proposed, contested)
- alongside: SOTA-133 — attention residuals     (Active, emerging)
```

## Checks

845 tests; `luria lint` clean here and on the consumer record, where [anthology#26](https://github.com/dmarx/anthology-of-the-sota/pull/26) runs all of it end to end and is green.


luria/config.py
```diff
@@ -930,27 +991,65 @@ def _references(prefix: str, raw: dict) -> tuple[Reference, ...]:
...
+        raise ValueError(f"{where}: the field is already always required, so "
+                         f"`required_when` says nothing — drop one of them")
+    on, values = next(iter(raw.items()))
+    values = tuple(str(v) for v in
+                   (values if isinstance(values, list) else [values]))
+    if not values:
+        raise ValueError(f"{where}: `required_when.{on}` lists no values, so "
+                         f"the condition can never hold and the field is "
+                         f"never required")
+    return RequiredWhen(on=str(on), values=values)
```
dmarx, line 1022: **A typo never fires, and nothing says so.** This validates shape only. `required_when = { staus = ["Proposed"] }` and `{ status = ["proposed"] }` are both accepted, never hold, and the field is never required — which is exactly the outcome the docstring above gives as the reason for validating eagerly ("a condition that can never hold … surfaces as no violations").

The scheme already knows what it would take to catch both: which fields it declares (its own `fields`, `references`, plus `status` and `tags`), the status vocabulary, and each vocabulary field's values. I'd refuse an `on` the scheme cannot name, and for `status` and vocabulary-typed fields, a value outside the set. A free-text field (`stage = ["blocked"]`) can't be checked and needn't be.

This is the one item I'd want before merge: the two most likely authoring mistakes currently pass silently, and the rest of the mechanism is built on the premise that they don't.

---

luria/config.py
```diff
@@ -233,6 +237,56 @@ class TagGroup:
...
+        `Proposed — pending a replication` as some other status and exempt
+        the document silently."""
+        raw = meta.get(self.on)
+        if raw is None:
+            return False
+        if self.on == "status":
+            from .statuses import parse
+            return parse(str(raw)).value in self.values
+        if isinstance(raw, list):
+            return any(str(v) in self.values for v in raw)
```
dmarx: **Normalisation covers status and nothing else.** The docstring says the compared value is normalised "the way the field's own machinery normalises it", and that's true for exactly one field. `holds` reads raw frontmatter, so a vocabulary field with a `default` — never absent by ADR-076's rule — reads as absent here: `required_when = { worlds = ["B"] }` with default `B` never holds for precisely the documents it means, silently.

Since the compiled contract already knows how to resolve a field's effective value (the same path the vocabulary check and the record page use), I'd have `holds` resolve `on` through that rather than special-casing `status` here. That also moves the `statuses.parse` import out of config, which otherwise reaches up into a module that depends on it.

Two smaller things the same function decides without the ADR saying so: a list-valued `on` (tags) matches on *any* element, and a missing `on` is "does not hold" rather than an error. Both are reasonable; both belong in the decision text.

---

luria/contract.py
```diff
@@ -57,8 +57,22 @@ class Field:
...
 
+    def demanded(self, meta: dict) -> bool:
+        """Whether this document must carry the field. Consulted everywhere
+        `required` used to be read directly, so a conditional requirement
+        cannot be honoured by one check and ignored by the next — which is
+        the failure mode `contract.py` was written to end."""
+        if self.required:
+            return True
+        return (self.required_when is not None
+                and self.required_when.holds(meta))
```
dmarx: **The built-in supersession rule is this exact shape and doesn't use it.** ADR-071's "a Superseded document must name `superseded_by`" is still a hand-written branch in `lint.check_frontmatter` (`elif status.value == "Superseded" and not status.superseded_by`), while this PR adds the mechanism that states it:

```python
Field("superseded_by", reference=ANY_SCHEME, many=True, builtin=True,
      required_when=RequiredWhen("status", ("Superseded",)), ...)
```

Doing that deletes a special case (DP-4: one implementation), gives the built-in the same finding wording, `because:` provenance and record-page line ("required when `status` is `Superseded`") for free, and — per the working agreement — fires the new mechanism on a real case in this record on day one, on the three Superseded decisions here. The only wrinkle is that `describe()` skips builtin fields, so the record page would need to decide whether a built-in conditional is worth a line; I'd say yes.

---

record/decisions.d/ADR-tmpngw57.md
```diff
@@ -0,0 +1,111 @@
...
+for every other: `statuses.yaml` is a vocabulary file, and making it a
+requirements file too gives it two jobs. It also puts the rule somewhere the
+person declaring the field would not think to look.
+
+**A general predicate language** — negation, conjunction, comparison across
+fields. Rejected. The entire value of this rule is that a reader sees it in
+the line that declares it; a config that can state arbitrary predicates is
+one nobody reads at a glance. One field against a set of literal values is
+enough for every case anyone has produced. A second real case can argue for
+more, which is a better way to learn the shape than guessing at it now.
```
dmarx, line 120: **Name the foreseeable second case here, so it isn't relitigated as the status-side placement.** The rule is per field. The moment a scheme has three fields that all apply while provisional (`promote_when`, `blocked_by`, `retracted_because`…), the same `{ status = ["Proposed", "Deferred"] }` is restated three times, and the three sets can drift — the DP-3 shape. Whoever hits that will reach for `statuses.yaml: Proposed: { requires: [...] }`, which this decision has just rejected for good reasons.

The answer that keeps the field-side placement is probably a `field_groups` entry that takes `required_when`, or a named condition declared once and cited by field. Neither needs building now; the ADR should say the repetition is expected and where the fix goes, in a sentence.

---

record/decisions.d/ADR-tmpngw57.md
```diff
@@ -0,0 +1,111 @@
...
+documents would carry a key with nothing to put in it, and a field that is
+usually empty is a field readers learn to skip.
+
+## What this does not do
+
+It cannot check that a condition is a *good* condition — that it names a kind
+of evidence rather than a quantity of it. Nothing mechanical can, and a check
+that tried would ratify bad conditions by passing them. What it does is make
+the condition exist, as a field, on the document, where the next person to
+touch that document reads it without needing the coincidence.
```
dmarx, line 160: **The inverse goes unsaid.** A `promote_when:` on a document that has since gone `Active` is a condition that outlived its purpose — the same shape as a directive that no longer matches anything, which this record already reports (`stale-directives`). #141 dropped `forbidden` because nothing would emit it; this mechanism is the first thing that would.

I wouldn't make it an error (the old condition can be worth keeping as history of *why* it was promoted), but a warning class on the dial fits ADR-035 exactly. If it's deferred, this section should say so and why — otherwise the next reader assumes it was overlooked rather than chosen.

---

dmarx (commented):
Review of the `required_when` part, which is the piece worth a careful read here. The shape is right and the ADR argues it well: on the field rather than the status, one field against a set of literals, `required` and `required_when` exclusive, one read path through `Field.demanded`, a finding that names the value that fired it, and the record page rendering the condition. The status normalisation with its test is exactly the kind of thing that gets missed.

Five inline comments. The first (load-time validation of the condition's field and values) is the one I'd want before merge — the mechanism's own justification for eager validation is the failure it currently lets through. The others are: normalise through the contract rather than special-casing `status`; express ADR-071's built-in supersession rule with this mechanism instead of a hand-written branch; name the repetition-across-fields case in the ADR so nobody reaches for the status-side placement to solve it; and say whether the inverse (a condition outliving its status) is deferred or overlooked.

Two things with no line to hang them on:

- `docs/modeling.md`'s rule table ("the rule you would write in prose → the table that enforces it") is where an adopter learns what the config can say, and `required_when` isn't in it.
- This PR carries three issues, two ADRs and ~2,400 lines, and the conditional requirement is the smallest of the three. Its files barely touch the chains work; if you want it reviewed on its own terms it would split out cleanly.

---

dmarx:
All five inline comments addressed and pushed; replies on each thread. Both no-line items below.

**The blocking one was the right one to block on.** `required_when` validated shape and not content, so a misspelled field and a miscased status were both accepted, both never held, and both left the field silently never required — word for word the failure the module's own docstring gives as the reason for validating eagerly. Validation now runs once the scheme is assembled, since a condition can name a field declared in another table.

**The vocabulary-default bug I would not have found.** Reading raw frontmatter meant a condition naming a field's own `default` never held, for precisely the documents it was written about. Resolution goes through the compiled contract now, which had the answer already.

**ADR-071 turned the feature into a deletion.** Stating the supersession rule as a `required_when` on the built-in removed the hand-written branch and the built-in inherited the finding wording, the provenance and the record-page mention. Fired on a real case in the consumer record.

One place I went the other way, on the thread: rendering the built-in in `describe()` contradicted the record page's own "nothing beyond the standard fields" sentence and made every scheme look as though it had declared a contract — a test caught it. The rule is stated once in that section's intro instead, with the standard fields. Intent kept, placement moved.

841 tests pass; `luria lint` clean here and on the consumer record.

---

### `docs/modeling.md`

Added to the rule table, which is exactly where I would have looked and didn't:

| a field is required only while the entry is unsettled | `fields` with `required_when = { status = ["Proposed", "Deferred"] }` | a violation naming the value that fired it |

### On splitting

You're right on the merits — three issues and two ADRs in ~2,400 lines is too much to review as one thing, and `required_when` is the piece that deserved its own read.

I'd rather not split it *now*, for one reason that isn't inertia: the review already happened, on this PR, and it found three real bugs. Splitting at this point re-opens a review that has landed and re-bases the other two on a moving base, and [anthology#26](https://github.com/dmarx/anthology-of-the-sota/pull/26) pins this PR's head SHA to demonstrate the whole set working together. The cost is now higher than the benefit it would have had two days ago.

What I'd take from it instead, and would rather commit to: **one issue per PR from here**, and where a PR adds a mechanism *and* uses it, the mechanism lands first. #169, #170 and #171 arrived together because the consumer record produced all three findings in one sitting — which is a reason they were discovered together, not a reason to ship them together.

Say the word if you'd still like `required_when` extracted and I'll do it; it does come out cleanly, and I'd re-point #26 at the new head.

---

dmarx:
split out

dmarx:
Split into three, commits cherry-picked as they were, each branch carrying no regenerated views (ADR-068):

- #174 — `template-drift` (#169), the `luria new` fixes, and the HTML-entity fix. Based on `main`.
- #175 — `required_when` (#170), including the follow-up commit that answered the review here. Based on #174's branch because the two compose in `luria/templates.py`; retarget to `main` once #174 merges.
- #176 — `[luria.chains]` (#171). Based on `main`; the only overlap with the others is one line in the lint's `FAILABLE` tuple.

Together the three reproduce this PR's sources file for file. Closing this one in favour of them.

---

Issue #160, opened 2026-09-05
A typed reference names exactly one scheme, so a relation with two legitimate target families cannot be declared


dmarx:
`Reference.scheme` is a single `str`:

```python
@dataclass(frozen=True)
class Reference:
    field: str
    scheme: str
    required: bool = True
    many: bool = False
```

`many` says the field holds *several codes*. Nothing says the field holds a code from *either of several schemes*. So a relation whose legitimate targets span two families cannot be declared at all, and the record has to either narrow it dishonestly or drop it into prose.

## The motivating case

`examples/constitution` declares:

```toml
[luria.schemes.BOUNDARY.references]
overrides = { scheme = "PRACTICE", many = true, required = false }
```

A boundary that overrides another **boundary** is rejected:

```
record/boundaries.d/BOUNDARY-003.md: `overrides: BOUNDARY-001` is not a PRACTICE code
  — a BOUNDARY document's `overrides` names a PRACTICE document
  (luria.toml: schemes.BOUNDARY.references.overrides)
```

That is not a contrived shape. Constitutions carry authorization carve-outs constantly — a permission that beats a prohibition — and the example already smuggles one in as prose, in `BOUNDARY-001`'s body:

> What it does not do is expand. A request that merely *sounds* like the class is not in it…

That sentence is doing the work an `overrides` edge to another boundary would do, unchecked, which is exactly the arrangement typed references exist to replace ([ADR-060](https://github.com/dmarx/luria/blob/main/record/decisions.d/ADR-060.md)).

The same shape shows up outside this example: an anthology where a practice may cite either a paper or a dataset; a spec that supersedes either a spec or an RFC.

## The workarounds, and why neither is good

**Two fields** — `overrides_practice` and `overrides_boundary`. The relation is one relation; splitting it by target type puts the type system's job in the field name, and every consumer of the graph now has to know to union two edges. It also reads as two different relations to anyone browsing the frontmatter.

**One loose field** — drop the reference, use `requires`. That is the failure [ADR-060](https://github.com/dmarx/luria/blob/main/record/decisions.d/ADR-060.md) was written against: `requires` is satisfied by any truthy value, so it checks presence and nothing else.

## Suggested shape

`scheme` accepts a list, and the check becomes membership rather than equality:

```toml
overrides = { scheme = ["PRACTICE", "BOUNDARY"], many = true, required = false }
```

[...]

PR #176, opened 2026-09-06
A relation is walked transitively and rendered as sequences (`[luria.chains]`)

dmarx:
Split out of #172 (third of three). Implements #171. Independent of the other two; the only overlap is one line in the lint's `FAILABLE` tuple, which whichever merges second resolves trivially. Commits are the originals, cherry-picked; this branch carries no regenerated views (ADR-068).

## `[luria.chains]` (#171)

Typed edges gave every page its neighbours; nothing answered *what sequence is this a step in*. So the sequences lived as prose, re-described once per participant, and two notes went stale on the same fact — each claiming two rival designs had never been compared — and were corrected in two places by luck.

```toml
[luria.chains.lineage]
scheme   = "LIT"
relation = "extends"              # the spine, directed
sibling  = "compared_against"     # optional cross-link, symmetric
output   = "docs/lineage.md"
```

The page renders order, title and status and nothing else: the field carries the sequence, the prose keeps the argument. A view that read well enough to replace the paragraphs would invite deleting them, and the paragraphs are where "why this step happened" lives.

`broken-chains` reports a succession that loops and a comparison only one side declares. Neither is acknowledgeable — there is no reading under which either is what the author meant.

Three things the implementation had to get right: a line is a DAG rather than a list (two papers can replace the same one), a rival that succeeds nothing still joins the component through its cross-link, and both graph walks terminate on a cycle. Two follow-ups from the first real corpus are here too: a chain's relation fields are not read as citation sites, and a step renders its status value rather than the composed display form.

Recorded as `ADR-tmpichk4`.

## Checks

- `python -m pytest tests -q`: 782 passed.
- `luria lint` on this branch, no view written: clean.
- As #172 said: the chain machinery has not met a real corpus in full yet — migrating one line and reading the result before converting the other nine is the next change.

https://claude.ai/code/session_01NCmBUY14BKhLGbHNKLuP6r

---
_Generated by [Claude Code](https://claude.ai/code/session_01NCmBUY14BKhLGbHNKLuP6r)_


dmarx:
Reviewing this as its first consumer rather than as its author — some of what follows is a criticism of code I wrote.

**Short answer on the Kuhnian question: it gets the topology right and the temporality wrong, and one of its checks actively fights the thing it was built for.**

## The check that has to go before this merges

`compared_against` is required to be declared on both sides, and a one-sided declaration is a `broken-chains` finding. I built that on the argument that a comparison is a fact both documents hold. **The walk does not need it.** `lines_of` already adds the edge in both directions:

```python
neighbours[code].add(other)
neighbours[other].add(code)
```

Tested on a clean fixture — three notes, only the comparing paper declaring anything:

```
one-sided declaration →
  members : ['LIT-001', 'LIT-002', 'LIT-003']
  findings: 2
  page has all three: True
```

The page is complete and correct. The two findings are demanding back-references that change nothing except the edit count.

What it costs in practice, in the record today: **three documents edited for one comparison.** LIT-152 declares two rivals, and LIT-134 and LIT-140 each had to be opened and edited to declare LIT-152 back. The fifth rival in a family means editing four existing notes to file one paper.

That is precisely the shape this feature exists to remove. The original bug was a fact restated in every participating note, going stale in two places at once; requiring symmetric declaration reintroduces it as field-editing churn instead of paragraph churn. It is also backwards against how comparison actually happens — a later paper compares itself to an earlier one, and the earlier one cannot possibly have declared a comparison to work that did not exist.

**Recommend: drop the symmetry finding.** One declaration, symmetric interpretation, which is already what the walk does.

## The rendering gap the Kuhnian picture needs most

A cross-linked rival renders flat at the bottom of the section, detached from the step it contends with:

```
- LIT-141 — Hyper-Connections                    (Superseded)
  - LIT-140 — mHC                                (Active)
    - LIT-151 — oHC                              (Active)
    - LIT-181 — sHC (spectral sphere)            (Active)
- alongside: LIT-134 — Attention Residuals       (Active)
- alongside: LIT-152 — Qwen3.8-Next              (Active)
```

LIT-134 contends with **LIT-140 specifically**, and the page does not say so. The `extends` forks nest correctly; the `compared_against` rivals do not attach anywhere. Since the fork point is the whole question — *where does the agreed part end and the contention begin* — this is the render that matters and it is the one that flattens.

The fix is small: nest a cross-linked document under the step it cross-links to.

## The deeper gap: no time axis

This is the one I would not have found without asking the Kuhn question, and I do not think it should block the merge — but it should be written down.

**Most scientific disputes end by attrition, not by refutation.** Nobody publishes "we were wrong"; people stop working on the losing branch. A chain page renders `contested` identically whether the last shot was fired last month or in 2021 — and those are opposite states. The four residual papers landed inside one month, which is what makes that fork live; had they been spread over four years it would be a fossil.

Two consequences:

- The page cannot show pace, so a reader cannot tell live contention from an abandoned branch.
- It is the concrete mitigation for the risk [ADR-015](https://github.com/dmarx/anthology-of-the-sota/blob/main/record/decisions.d/ADR-015.md) flags in the consumer record: a consensus reading ages with nothing to trigger a re-read. A line whose newest node is three years old is *self-evidently* stale, and the page has the dates and does not show them.

Related, and also not expressible: **a fork that closed.** `superseded_by` retires a document by naming its successor. A branch that lost by attrition has no successor to name — normal science resumed and nobody wrote it down. Retiring it would be wrong (it may still be live elsewhere) and leaving it `contested` forever is also wrong.

## Smaller: the two layers don't join

The framing that prompted this was *a lineage chain up to a commonly agreed SOTA practice from which competing experimental extensions fork*. That is a **practice** trunk with **paper** forks, and chains are per-scheme, so it renders as two pages a reader joins in their head. Related to #160.

## What actually works, and it is the bulk of it

The spine, the DAG handling, the cycle guard, the components, the "field carries the sequence, prose carries the argument" split — all of that holds up on a real corpus. The four bugs the first real record found are fixed. `alongside` was the right *concept*; only its placement is wrong.

## Suggested disposition

- **Before merge:** drop the `compared_against` symmetry finding. It trains a habit that is the opposite of the point, and it is one deletion.
- **Small follow-up:** nest cross-linked rivals under the step they contend with.
- **Issues worth filing:** a date on each step, and a way to say a fork closed without naming a successor. I'll write both up if you want them.

---

Issue #181, opened 2026-09-06
`status:` should be an ordinary controlled vocabulary, not a built-in axis


dmarx:
ADR-076 generalized the controlled-vocabulary mechanism **from** `status` and `tags`, used *"`status` and `tags` are themselves vocabulary-backed fields"* to reject an alternative design, and then carved both out:

> The built-in axes — `status`, `tags` — cannot be redeclared here; they have their own files and their own rules.

So a bespoke reimplementation runs beside the general one. Recorded as `ADR-tmpv6jcf` on `claude/statuses-are-the-projects-e6evzh`.

## The duplication is observable, not theoretical

Wire a scheme's `status:` to a vocabulary and give a document a bad word, and you get **two findings for one value**:

```
VP-002.md: status 'Bogus' is not one the VP scheme uses — want Active|Withdrawn …
VP-002.md: `status: Bogus` is not in the `statuses` vocabulary (…) — the values are Active, Withdrawn
```

Same shape elsewhere: `_check_conditions` has a hand-written `if when.on == "status"` branch immediately beside the generic `elif when.on in vocab_of` that would do the same job.

## What it costs today

`statuses.CLOSED` made the vocabulary a law — `statuses.yaml` could only *narrow* the five. A project whose decisions are `Accepted` and `Withdrawn` cannot say so; one that retires them into `supplanted_by:` cannot either. Nothing generic needs those words: the retired-citation check reads only `scheme.active` and treats everything else as out of force, and remotes check existence, never status.

## Proposed

```toml
[luria.schemes.ADR.fields.status]
vocabulary = "statuses"
```

- `statuses.yaml` becomes an ordinary vocabulary file; the bespoke reader and validation are deleted.
- **The absence of the declaration is an error, not a silent fallback to the five.** A default inherited from code cannot be read, so nobody learns it is theirs to change — that is the `template-drift` / `inert-status` failure (live, enforced, invisible) and this should not add a third instance.
- **The break ships with the fix:** `luria upgrade statuses` writes the declaration and the file into an existing project. It must **not** go through `load()`, because the config it repairs is the config that now refuses to load. It is temporary by construction and carries a marker so it gets removed rather than accumulating.

## Verified

- Removing `status` from `BUILT_IN_AXES` breaks nothing — 888 tests still pass. The carve-out is a wall, not a dependency.
- A spiked project with the declaration validates statuses correctly through the generic path.

## Three known obstacles

[...]

Issue #173, opened 2026-09-06
A chain renders an agreed trunk and a disputed fork identically: let a chain annotate its steps with a declared field


dmarx:
`[luria.chains]` (#171) renders each step as code, title and status. That is
the right floor, and it turns out to be one field short for the question a
line of work is most often read to answer: **which part of this is agreed, and
where does it fork?**

## The case

The consumer record just added a second axis to its practices — how far the
*field* has converged, as distinct from what the record itself asserts. The
two are orthogonal, and the interesting entries are the ones where they
disagree: a design shipped in a 1.6T-parameter production model, attacked by
two independent groups within a month, and still `Proposed` in the record.

The chain page had that in its data and could not show it:

```
- SOTA-137 — freely learned mixing                (Superseded)
  - SOTA-136 — doubly-stochastic mixing           (Proposed)
- alongside: SOTA-133 — attention residuals       (Active)
```

The middle step is the one under active dispute and it renders like any other.

## Proposal

```toml
[luria.chains.practice]
scheme   = "SOTA"
relation = "extends"
output   = "docs/practice-lines.md"
annotate = "consensus"
```

```
- SOTA-137 — freely learned mixing                (Superseded, unassessed)
  - SOTA-136 — doubly-stochastic mixing           (Proposed, contested)
- alongside: SOTA-133 — attention residuals       (Active, emerging)
```

Three things it should get right, each with a reason rather than a preference:

- **Read through the contract, not the frontmatter.** A vocabulary field with
  a `default` is never absent (ADR-076), so a step that says nothing still has
  a value; reading raw would make the page disagree with the record. Same bug
  as the one `required_when` shipped with in #172.
- **A field the scheme does not declare is a config error.** Every step would
  render blank, and blank looks exactly like correct (DP-15) — the same
  argument as a chain over an undeclared relation.
- **One field, not a list.** A step's line is already long. If a second
  annotation is ever wanted, that is a real case to argue from.

Deliberately *not* proposed: rendering the annotation as anything but text
beside the status. Grouping or sorting a chain by the annotated field is a
different feature, and the value of this one is that the fork becomes visible
without the page taking a position on it.

dmarx:
Shipped on `main`; closing.

[...]

PR #187, opened 2026-09-06
A chain can annotate its steps with a declared field (#173)

dmarx:
Implements #173. This was written while #175 and #176 were in flight — it needed `Contract.reading` from one and the chain renderer from the other — and has been parked since. Both are merged, so here it is.

## Why

A chain page renders order, title and status. For a line of *practice* that is not enough. The fork the consumer record argues about is a `Superseded` trunk with a `contested` branch beside an `emerging` sibling — and with one axis those three read identically:

```
- SOTA-137 — …freely learned mixing (hyper-connections)   (Superseded)
  - SOTA-136 — …doubly-stochastic matrices (mHC)          (Proposed)
- alongside: SOTA-133 — …attention over preceding layers  (Active)
```

Nothing there says which branch the field is arguing about.

## What

```toml
[luria.chains.practice]
scheme   = "SOTA"
relation = "extends"
annotate = "consensus"     # a second declared field, shown beside status
```

```
- SOTA-137 — …freely learned mixing (hyper-connections)   (Superseded, unassessed)
  - SOTA-136 — …doubly-stochastic matrices (mHC)          (Proposed, contested)
- alongside: SOTA-133 — …attention over preceding layers  (Active, emerging)
```

Two things it gets right:

- **Read through the compiled contract, not the raw frontmatter.** A field with a `default` shows its default rather than a blank — what ADR-076 means by an effective value never being absent. `unassessed` above is a default nobody wrote.
- **Validated at load** against what the scheme can actually name. An `annotate` naming nothing renders every step blank, and blank is indistinguishable from correct (DP-15).

## One thing that changed while it was parked

The original carried `{"status", "tags"}` as hardcoded names in that validation set. After #181, `status` is a declared vocabulary and arrives through `vocabularies` like any other field, so naming it there is the built-in-in-code habit that PR was about. Dropped; only `tags` remains, which is still an axis the code assumes.

## Checks

- `python -m pytest tests -q`: **904 passed**.
- `luria lint`: clean.
- **Fired on the real corpus.** The anthology needed `luria upgrade statuses` first — #184 landed while this was parked, which is the round trip I flagged on that PR — and after it the practice chain renders `(Superseded, unassessed)`, `(Proposed, contested)`, `(Active, emerging)`. The branched conflict legible in one page, which is what the second axis was added for.

This also unblocks the consumer: [dmarx/anthology-of-the-sota#25](https://github.com/dmarx/anthology-of-the-sota/pull/25) has `annotate = "consensus"` in its config already, ignored by 0.9.0, so a release carrying this restores the annotation there with no change on that side.

---

### DIFF
--- a/luria/chains.py
+++ b/luria/chains.py
@@ -219,6 +219,17 @@ def _link(doc: Adr, chain) -> str:
     return prefix_for(scheme, chain.output.parent) + doc.path.name
 
 
+def _annotation(doc: Adr, chain) -> str:
+    """The chain's second axis for one step, read as the contract reads it —
+    so a field with a `default` shows its default rather than a blank, which
+    is what ADR-076 means by "never absent"."""
+    if not chain.annotate:
+        return ""
+    obligations = for_scheme(current().schemes[chain.scheme])
+    values = obligations.reading(chain.annotate, doc.meta)
+    return f", {', '.join(str(v) for v in values)}" if values else ""
+
+
 def _step(doc: Adr, chain, lead: str = "") -> str:
     """One line of the rendered list: the code, the title, and the status
     *value*.
@@ -233,7 +244,7 @@ def _step(doc: Adr, chain, lead: str = "") -> str:
     to want. `status_value` is the field, and the fields are why it is
     there to ask for."""
     return (f"{lead}[{doc.code}]({_link(doc, chain)}) — {doc.title} "
-            f"*({doc.status_value})*")
+            f"*({doc.status_value}{_annotation(doc, chain)})*")
 
 
 def _render(chain, lines: list[Line]) -> str:
--- a/luria/chains.py
+++ b/luria/chains.py
@@ -221,7 +221,7 @@ def _link(doc: Adr, chain) -> str:
 
 
 def _annotation(doc: Adr, chain) -> str:
-    """Every field this chain shows for one step, in the order it names them.
+    """One step's value along each facet the chain names, in that order.
 
     Read through the compiled contract rather than raw frontmatter, so a
     field with a `default` shows its default rather than a blank — what
@@ -230,7 +230,7 @@ def _annotation(doc: Adr, chain) -> str:
     obligations = for_scheme(current().schemes[chain.scheme])
     meta = statuses.normalised(doc.meta)
     shown: list[str] = []
-    for field in chain.annotate:
+    for field in chain.facet_by:
         shown += [str(v) for v in (obligations.reading(field, meta) or ())]
     return ", ".join(shown)
 
--- a/luria/chains.py
+++ b/luria/chains.py
@@ -51,6 +51,7 @@ import re
 from dataclasses import dataclass, field
 
 from .adr_index import Adr, load_scheme, prefix_for
+from . import statuses
 from .config import current
 from .contract import for_scheme, values_of
 from . import relations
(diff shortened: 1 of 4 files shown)


dmarx:
```
[luria.chains.practice]
scheme   = "SOTA"
relation = "extends"
annotate = "consensus"     # a second declared field, shown beside status
```

I think I'm missing something, that code snippet doesn't include any mention of status, so it's unclear how that relationship is being parameterized if "status" is now a declared vocabulary like any other and not a special case built-in.

dmarx:
You're not missing anything — that was the bug, and the snippet was the tell.

The chain didn't mention `status` because **the renderer hardcoded it**:

```python
return (f"{lead}[{doc.code}]({_link(doc, chain)}) — {doc.title} "
        f"*({doc.status_value}{_annotation(doc, chain)})*")
```

`annotate` meant *"the other field"* — status rendered first, unnamed, always. Which left the chain page as the last thing in the codebase treating `status` as a built-in axis, a day after #181 made it a declaration. Exactly the thing that PR was about, hiding one layer down.

Fixed in `19683c3`: **`annotate` is the whole list, in render order.**

```toml
annotate = ["status", "consensus"]   # or ["consensus"] alone, or omitted
```

- **Defaults to `["status"]`**, so a chain that says nothing renders exactly what it always rendered.
- A scalar reads as a list of one.
- Every named field is checked at load against what the scheme can name — a field it can't name renders blank on every step, and blank is indistinguishable from correct (DP-15).
- `status` is nameable because it now arrives through `vocabularies` like any other declared field. Only `tags` is still assumed.
- Reading goes through the compiled contract, so a `default` shows and a `status:` carrying a note reads as its word.

There's a test named for the property, since this is the kind of thing that creeps back:

> **`test_annotate_names_every_field_shown_including_status`** — asserts `annotate = ["consensus"]` renders the consensus value and *no* status.

908 tests, lint clean. Re-verified on the anthology after `luria upgrade statuses`: still `(Superseded, unassessed)`, `(Proposed, contested)`, `(Active, emerging)` — now because the config names both fields rather than because the code assumed one.

One consequence for the consumer: its `annotate = "consensus"` becomes `annotate = ["status", "consensus"]`, or it loses the status column. One line, and I'll make it in the same pass as the pin bump whenever you cut the release.

---

dmarx:
Your instinct was better than my first answer. `9c2e0dd`:

```toml
facet_by = ["status", "consensus"]
```

## Why the old word stopped working

`annotate` was accurate for a design that no longer exists. It meant *"the OTHER field"*, beside a status the renderer hardcoded — a verb whose object was the thing it sat next to. When the fix made the list the whole content, there was nothing left for it to be an annotation **on**. The word outlived the shape it was named for.

So: a weakness in the language, and the ergonomics were only wrong for as long as the hardcoded status was.

## Why not `show`

That was my first move and it's the wrong kind of word — it names what the *renderer does*, not what the *fields are*. Any config key could be called `show`. It describes the mechanism and says nothing about why `status` and `consensus` belong in one list.

## Why `facet`

These are **independent axes along which one document is classified** — which is what a facet is, and the concept was already in the record before the key was. ADR-076 calls `status` and `tags` "the built-in **axes**"; `BUILT_IN_AXES` is literally the constant name; and the vocabulary machinery already writes a page per value, which *is* faceted browsing. The anthology's own ADR-015 introduced `consensus` as "a second axis, orthogonal to `status`".

So the word isn't imported — it's the one this record was already using in prose.

`index` was the other candidate and collides twice: `render = "index"` and the `luria index` command.

`facet_by` rather than `facets` or `facet` because it's verb + preposition, the shape `required_when` and `retires_on` already use, and it reads as the instruction it is: *facet by status and consensus*.

One thing I'll flag rather than let you find it: faceting usually implies you can **filter** by the axes, and the chain page only *displays* them. I think the word still earns it, because ADR-076's page-per-value already provides the filtering half and this shows the same axes per step — but if you read that as overreach, `facets = [...]` as a plain noun is a one-line change.

908 tests, lint clean, re-verified on the anthology: `(Superseded, unassessed)`, `(Proposed, contested)`, `(Active, emerging)`.

---

Issue #216, opened 2026-09-08
A field can be derived from another field


dmarx:
## The shape

```toml
[luria.schemes.SOTA.fields.primary_topic]
derive     = "first:tags"
vocabulary = "tags"
```

`primary_topic` is then an ordinary field everywhere a field is read — `invariant`, `facet_by`, `required_when`, reports, the site — and it is never written down.

## What it is for

Today "primary topic" is not a field. It is a **tag group with `require = "exactly-one"`** — a constraint on a set. Asking a document for its primary topic means intersecting its tag set with the group's and hoping one value falls out, which is why the whole model is order-blind and why *holding two vocabulary topics is a violation rather than a statement*.

That constraint has a cost that showed up the first time #214's check ran on a real record. All 12 unbound relations it found pair two documents whose topics differ, and **not one of them can be fixed by naming the topic they share** — `exactly-one` forbids the second topic. The only legal remedies are inventing a new vocabulary word or attaching a private tag. Both were used in dmarx/anthology-of-the-sota#101; neither is what the record actually wanted to say.

Order is already the vocabulary file's own convention — the anthology's `tags.yaml` opens with *"in the order the index shows them"* — so reading `tags[0]` as the primary is picking up a convention that already exists rather than inventing one.

With the derivation:

- `exactly-one` relaxes to "the first tag is primary-eligible", and a document may carry two real topics.
- `invariant = "tags"` gains teeth, because a shared topic becomes expressible.
- `invariant = "primary_topic"` becomes available as the strict reading, alongside the loose one.

## Not the URL machinery

`Remote.uri` renders a URL from a curated template vocabulary keyed on remote + code. It formats strings; it does not compute a document's field from that document's other fields, and nothing in it is reusable here.

What *is* relevant is its policy. [ADR-066](https://github.com/dmarx/luria/blob/main/record/decisions.d/ADR-066.md) settled that **a declaration beats a derivation** for URIs. This proposes the opposite for fields: **a derived field is read-only, and writing one is a violation.** A URI override is a genuine second source of truth for a thing luria cannot otherwise know; a written `primary_topic:` is the *same* fact as `tags[0]`, stored twice, free to drift — which is the failure ADR-060 already documented for this exact vocabulary (seven terms in four places, two copies disagreeing). The departure is deliberate and belongs in an ADR.

## Scope: a closed set, not an expression language

[...]

Issue #214, opened 2026-09-08
Report a lineage path that no tag unifies


dmarx:
> Rewritten after [this comment](https://github.com/dmarx/luria/issues/214#issuecomment-5587259471). The first draft framed the defect as a lineage *crossing a topic boundary*, which treats overlapping membership as a problem when it is a feature. The real signal is the absence of a unifying tag. Reasoning for the change is in the thread.

## The check

A record with typed relations and tags has two ways of saying "these documents belong together", and nothing compares them.

- A **lineage** is a set the record has *asserted* belongs together — connected via `extends`, `corrects`, `compared_against` and their converses.
- A **tag** is a name for something documents have in common.

If the record asserts a path, it believes those documents share something. **If no single tag is held by every member of that path, the record has not said what.** That is the finding, and the usual reading of it is that the vocabulary is missing a word.

Concretely: take connected components over relation edges within a scheme, intersect the members' tag sets, report the components whose intersection is empty.

## Why not "the path crosses a topic boundary"

Because a document legitimately sitting in two areas is the vocabulary working, not failing. Divergence is not the defect; silence is.

The two checks agree on the anthology's current corpus — both flag one of twelve multi-member components — so that data does not choose between them. It cannot: no multi-member lineage there carries a shared *secondary* tag, which is the only place the difference shows. Constructed, it shows immediately:

```
before: primaries=['attention-techniques', 'model-stability']  shared=(none)
        crossing fires: True    unbound fires: True
after : primaries=['attention-techniques', 'model-stability']  shared=['kernel-numerics']
        crossing fires: True    unbound fires: False
```

Same lineage, primaries still divergent, but now something binds them. The crossing check keeps firing — a false positive silenced only by a directive or by forcing a document out of a topic it belongs in. The unbound check clears, because the record has now said what they have in common.

Strictly weaker, and weaker in exactly the cases where the stronger version was wrong.

## Why it earns its place

The anthology added three practice topics this week. Two came from grepping titles for `inference`, `serving`, `fine-tune` — which works, and finds only the categories you already suspect, because you have to name one to search for it.

The third came from this check, which names nothing. Of twelve multi-member components, three had no unifying tag:

| path | what it was |
|---|---|
| RoPE → rescale-for-context → drop-from-global-layers | a missing topic — one argument about positional encoding, five practices once fully counted |
| GRPO baseline → its correction → the alternative it's compared against | **a regression introduced by the retagging commit itself**, which moved one member and left two behind |
[...]

PR #217, opened 2026-09-08
A field can be derived from another field (#216)

dmarx:
Implements #216.

```toml
[luria.schemes.SOTA.fields.primary_topic]
derive     = "first:tags"
vocabulary = "tags"
```

`primary_topic` is then an ordinary field everywhere a field is read — the invariant a chain asserts, a facet, a report column — and is never written down.

## Why it exists

"Primary topic" was not a field. It was a `tag_group` with `require = "exactly-one"` — a constraint on a *set*. Asking a document for its primary meant intersecting its tags with the group's and hoping one value fell out, which is why that model is order-blind and why **holding two vocabulary topics read as a violation rather than as a statement**.

All twelve unbound relations #214 found on the consuming record pair documents whose topics differ, and not one can be fixed by naming the topic they share. The remedies left were widening the vocabulary or attaching a private tag; dmarx/anthology-of-the-sota#101 used both.

## Read-only, and deliberately unlike ADR-066

`ADR-066` settled that a declaration beats a derivation for a remote's URIs. This inverts it, and the ADR says why: a `url` template is a genuine second **source** for something luria cannot otherwise know, while a written `primary_topic:` is the same fact as `tags[0]`, stored twice and free to drift — the failure `ADR-060` already documented for this exact vocabulary. Override where you would otherwise be guessing; refuse where you would otherwise be duplicating.

## It composes rather than adds

Two properties of code already here kept this to ~150 lines:

- **`fields` was already "the table a field's shape and type are declared in"**, so `derive` pairs with `vocabulary` — and *"the first tag must be a real topic"* falls out of `_vocabulary_violations` with **no new check written**, so there is no second check to keep in step.
- **Frontmatter edits are text surgery, never a YAML round-trip** (`field_edit`). Tag order survives every tool luria has, so order-as-semantics is safe by construction. I checked this *before* designing anything — if any tool sorted tags the whole idea would have failed silently and late.

## Refused at load

`take` outside `{first, last}`; a source the scheme doesn't declare; a source that isn't list-valued (`first:` of one value is that value — a rename wearing a derivation's clothes); `many`, `required` or `default` alongside `derive`. `luria new` offers no flag for a derived field, since scaffolding one would write a line the lint rejects on the document's first read.

## Fired on the real case, twice

Per the working agreement, on the consuming record rather than on fixtures:

- **406 documents, every one already carrying a real topic in first position.** The convention existed; nothing enforced it and nothing could read it. Adoption is a no-op and the lint held its exact baseline.
- **Both guards made to fire on real documents** — a `primary_topic:` line added to a practice, and a `tiny-models` secondary moved to first position on another.

The second is the one a fixture would have let me get wrong: my first attempt tested it against a tag that only exists on an unmerged branch, so the check silently did nothing and looked like it passed. That is in the devlog.

## The cost, stated

Ordering carries meaning, and `tags: [a, b]` vs `[b, a]` look alike to a skimming reader. Nothing reorders, `luria new` writes the primary first, and a vocabulary-backed derived field catches an ineligible tag in first position — but **none of that catches two *eligible* topics written in the wrong order**. The ADR says so rather than leaving it to be discovered.

## Diff

| | |
|---|---|
| `luria/derive.py` | new — the derivation |
| `luria/config.py` | `Scheme.derived`, `derive` in `fields`, eager validation |
| `luria/contract.py` | derived before every check; writing one is a finding; `describe` says where the value comes from |
| `luria/adr_index.py` | resolved onto `Adr.meta`, the seam every reader shares |
| `luria/new.py` | no flag for a derived field |
| `tests/test_derive.py` | 22 tests |

**983 tests pass.** `luria lint` at `main`'s baseline, read whole — the single added line is this contribution's own `Proposed` decision. ADR, changelog fragment and devlog entry filed with the work; sources only, no views.

## Noticed, not fixed here

The anthology's `record/practices.d/tags.yaml` still cites `ADR-tmplvh9j` in a comment — a temporary code concretized to `ADR-021` on merge. `luria concretize` doesn't sweep vocabulary files and the reference checker doesn't read them, because they aren't documents. Worth its own issue.

---


dmarx:
I'm wondering if field derivation and URL construction could be unified under common machinery, e.g. jinja templating, JSONata, etc.?

dmarx:
Went and looked rather than answering from the design. Short version: I don't think they unify, and the reason is a type difference rather than a taste one — but the question surfaced something worth recording either way.

## How much duplicated machinery is actually there

Less than it looks. `Remote._render` is ~15 lines: build a `values` dict from the remote and the code, `template.format(**values)`, `KeyError → ""`. `derive.value` is ~5: read a list off `meta`, index it. There is no shared code to extract — unifying would be *creating* machinery, not removing it.

And luria isn't accumulating mini-languages. Counting every "computed from a declaration" in the codebase — `Remote.uri`, journal paths from `created:`, `Site.title`/`base_url` from `issue_url`, `Adr.code` from prefix+number, tag-group members from `primary_for`, and now `derive` — **exactly one has a user-facing expression surface**, the URI template vocabulary. The rest are fixed Python with nothing to write. A general engine would be the *first* general engine, and it would be justified by two members.

## The type difference

This is the part I'd defend hardest:

- **URL construction synthesizes a new string.** That is its job — `{repo}`, `{ref}` and `{anchor}` exist to be concatenated into something that did not previously exist.
- **Field derivation selects an existing value.** `first:tags` returns *an element of the list*, which is what preserves identity with the vocabulary. That identity is load-bearing: the derived value gets set-intersected against another document's tags by `invariant`, and checked against the closed vocabulary by `_vocabulary_violations`.

Under a string template, `{tags[0]}` happens to work and `topic-{tags[0]}` produces a value no vocabulary can ever contain — trivially writable, and silent. A template engine would make the always-invalid derivation easy to write, which is the exact failure the eager checks exist to prevent, reintroduced across a much bigger surface.

There's also a shape limit: a string engine can only ever return a string, so `many = true` derivations ("the tags that are also topics") stay impossible. That's latent today since I forbid `many`, but a unification would bake it in.

## The policies are opposite on purpose

`ADR-023`/`ADR-066` give URIs a four-rung precedence ladder where an explicit declaration wins. This ADR gives derived fields exactly one rung, read-only, and argues why. Common machinery would have to carry both as a flag — which is the unification buying nothing and then having to document that it means two different things depending on which caller you are.

## On the specific candidates

**Jinja** brings control flow — `{% if %}`, filters, loops — which is precisely the expression language the ADR rejects on the project's own "a mechanism joins when a case demands it" rule. It's also string-typed, so it inherits the problem above, and it'd be a third runtime dependency for a package that currently has two (`PyYAML`, `fire`).

**JSONata** is the better call of the two and I want to be fair to it: it returns *values* rather than strings, so it clears the type objection, and it can express list results. What it costs is a whole query language to document, support and version forever — plus a third-party Python implementation to depend on — in exchange for replacing a five-line function. And it's a poor fit for the URI half: URL construction isn't querying a document, it's filling a template, so JSONata would unify by making one of the two cases awkward.

## What I think is actually shared

Not machinery — **policy**. Both mechanisms hold the same rule: *a declaration that could never resolve is refused where the config is read, never discovered as silence.* That rule is currently stated twice in prose and implemented twice, and it's the thing a future third case should inherit. That's a design principle, not a base class.

## What would change my mind

A concrete third derivation that is **string-shaped and reads frontmatter** — `slug` from `title`, `year` from `date`. Two of those and `first:tags` becomes the odd one out rather than the archetype, and a template engine starts paying for itself. I'd rather wait for that case than guess at it.

Happy to be overruled — say the word and I'll do it. Otherwise I'd like to file this as an issue so the reasoning is findable when the third case shows up, rather than leaving it in a PR thread.

---

dmarx:
Withdrawing most of the comment above. Design captured in #218.

The count I led with — "exactly one user-facing expression surface" — was circular: it inventories a codebase whose constraints forbid the thing, finds it absent, and reports the absence as evidence of no demand. And you're right that the linter answers the string-typing objection; I built that guard in #216 and then argued as though it didn't exist.

Checked the historical claim against the anthology's frozen pre-migration YAML, and it's exact:

```yaml
- id: MLR-2014-Kingma001-0001
  topic_id: optimization/default-choice-for-neural-network
```

Both became `SOTA-001` and nothing.

`str.format` — already what `Remote._render` uses — does this today: `'LIT-{authors[0]}-{year}'` → `'LIT-Kingma-2014'`, `'{tags[0]}'` → `'model-stability'`. So the unification is **one template vocabulary, two variable sources** (remote+code for URIs, frontmatter for documents) plus a small filter set, with no new dependency.

One thing I'd still argue, and it changes the shape rather than the answer: derived **codes** make identity a function of a mutable field, and three attributions were corrected after filing in this collaboration alone. Derived **aliases** don't — and `aliases.py` already maintains exactly the many→one resolution map they'd need. Details in #218.

---

Issue #219, opened 2026-09-08
Plan: identity in the document, expressive aliases, one template vocabulary


dmarx:
The consolidated plan. Supersedes the scattered design in #216, #217 and #218; those stay as the trail.

> **Done.** All three steps are on `main`: #220 (`ADR-087`), #221 (`ADR-088`), #223. What follows is the plan as written, with the corrections it picked up along the way recorded in the comments.

## The goal

Recover what adopting luria cost the anthology. Its pre-migration identifiers, still frozen in `data/research.yaml` as provenance:

```yaml
- id: MLR-2014-Kingma001-0001
  topic_id: optimization/default-choice-for-neural-network
```

Prefix, year, first author, disambiguator, sequence — readable without a lookup — plus a slug. Both became `SOTA-001` and nothing. That was a real trade, and the goal here is to get the readability back without giving up the stability that made numeric codes worth trading for.

**Outcome:** `LIT-Dao-2022-074` resolves wherever `LIT-074` does, and `luria link --fix` leaves it written. Verified on all 218 literature notes, no collisions.

## How this got here, briefly

I argued against unifying field derivation with URL construction, on a count of "how many expression surfaces does luria have." That count was circular — it inventoried a codebase whose constraints forbid the thing, found it absent, and reported the absence as evidence of no demand. Demand had already manifested in URL construction, semantic codes/aliases, filename slugs, primacy inference, and site layout. The reasoning is in #218.

## Three rules, decided

**1. A derived value may participate in identity only if it is recomputed; it may only be recomputed if it isn't identity.** Mutually exclusive, so identity is never derived-on-read. Aliases are recomputed and therefore always true. Codes are assigned.

Evaluating a template once inside `concretize` and freezing the result is *technically* safe — identity never moves. It is still rejected: a seeded code **looks derived and isn't**, so correcting `authors[0]` leaves a code reading `Askell` forever, correctly-by-design, and every future reader has to be told. A numeric code makes no claim that can rot.

**2. A counter, not a UUID.** The counter carries ordering — the ADR scaffold says so: *"Numbering is sequential and carries information (it's the order decisions were made)."* A UUID discards that to buy merge-safety, and temporary codes (`ADR-049`) already solved merge-safety.

**3. Collisions get an allocator rather than a prohibition.** `concretize` owns a deduplicating field; a template that includes it is unique by construction. Order-dependence under auto-suffix is a cost the person choosing the template accepts.

---

## Step 1 — identity moves into the document ✅ #220

[...]

PR #220, opened 2026-09-08
Identity is a field; the filename is a projection of it (#219)

dmarx:
Step 1 of #219. A scheme document carries `uid:`; its filename derives from it.

## This is the journal model, applied to schemes

`check_journals` already says the whole thing out loud:

> *"A journal entry's path is derived from its `created:` timestamp, and the two have to agree — otherwise the ordering the whole scheme rests on says one thing and the frontmatter says another (ADR-020)."*

Schemes kept identity in the filesystem instead, and nothing anywhere says why. **So this is less a feature than the end of an inconsistency**, and most of the implementation is the journal's, moved one directory over.

| | |
|---|---|
| `Scheme.uid_of()` | reads the field, falls back to the filename |
| `luria lint` | reports a field and a filename that disagree |
| `luria repair` | writes the field from the path where it is absent |
| `luria new` / `luria concretize` | write it on filing / at allocation |

The fallback is what lets this arrive **without an upgrade anyone has to run**: a record written before the field existed reads exactly as it did, and acquires the field on the next `luria repair`.

## Why it matters beyond tidiness

Identity in the filename is what blocks expressive filenames and derived aliases. Anything a name encodes becomes load-bearing once the name *is* the identity, so a slug can't be added without making the slug part of what a citation resolves through. The anthology paid this: `MLR-2014-Kingma001-0001` plus a `topic_id` slug became `SOTA-001` and nothing.

## Three things already in the tree made it cheap

- **`documents()` is "the one place a scheme directory is read."** DP-4 consolidated five copies of that glob for unrelated reasons — so twenty-odd identity readers meant exactly one function to change.
- **`number_of()` already parses a trailing slug.** `adr-010-some-title.md` reads today; luria could always *read* expressive filenames and has only ever *written* the short form.
- **`journal.populate_created()` is the back-fill**, working, with a slot beside it in `repair.apply()`.

## Two judgement calls, and why

**The cache is keyed on `(mtime, size)`, not reset by hand.** Reading the field costs a parse on the hot path. The obvious cache is a module dict with a `reset()`, the way `aliases.py` does it — and that would have been a bug generator: `concretize` writes then re-reads, `repair` writes hundreds, and every future writer would have to remember. Keyed on the stat, a write invalidates its own entry.

**A disagreement is a violation, not a repair.** Renaming the file to match the field is tempting since the field is the truth — but which of the two is right is genuinely unknowable, and renaming on a guess moves identity and orphans links. An *absent* field has one witness (the path) and is repaired; a *contradicted* one has two and is reported.

## Fired on both records, per the working agreement

- **436 documents** in the anthology and **103** here acquired the field.
- `luria lint` held its **exact baseline** on both.
- **Every generated view came out byte-identical**, except `reference-status.md`, whose citation line numbers shift down by one — because the field adds one line, which is the answer you want that check to give.
- Then the guard itself on a **real document rather than a fixture**: `uid:` edited to disagree with its filename, reported, reverted.

## Diff

| | |
|---|---|
| `luria/config.py` | `Scheme.uid_of()`, `_declared_uid()`, the stat-keyed cache |
| `luria/lint.py` | `check_uids()` |
| `luria/repair.py` | `populate_uids()` — the migration |
| `luria/new.py` | `write_uid()`, written on filing |
| `luria/concretize.py` | written at allocation |
| `luria/adr_index.py`, `edges.py`, `site.py` | the three direct `number_of` readers |
| `tests/test_uid.py` | 19 tests |
| `record/**` (103 files) | the migration, run here rather than left to a bot commit |

**981 tests pass.** Lint exit 0, read whole — the two lines differing from `main` are this contribution's own `unlinted-file:` directive and its own `Proposed` decision.

## Worth watching

The new suite tripped the reference checker twice (`ADR-007` is Superseded here; `ADR-tmpabcde` resolves to nothing), because a suite about how codes are read has to write codes down. `tests/test_migrations.py` hit this before me and carries the same `unlinted-file:` directive. Two files isn't a pattern; three would be, and the fix then is for the scanners to know a test fixture is a quote rather than an address.

---

### DIFF
--- a/record/decisions.d/ADR-001.md
+++ b/record/decisions.d/ADR-001.md
@@ -1,5 +1,5 @@
 ---
-uid: 1
+number: 1
 status: Active
 title: 'Four layers of record, each with a test for what belongs in it'
 version: 2
--- a/record/decisions.d/ADR-001.md
+++ b/record/decisions.d/ADR-001.md
@@ -1,4 +1,5 @@
 ---
+uid: 1
 status: Active
 title: 'Four layers of record, each with a test for what belongs in it'
 version: 2
--- a/luria/adr_index.py
+++ b/luria/adr_index.py
@@ -155,9 +155,9 @@ class Adr:
         self.path = path
         self.scheme = scheme
         self.prefix = scheme.prefix
-        # Identity is the document's own `uid:` where it has one, the
+        # Identity is the document's own `number:` where it has one, the
         # filename's number where it doesn't (#219).
-        self.number = scheme.uid_of(path)
+        self.number = scheme.number_of(path)
         # A merge-allocated document awaiting concretization (ADR-049): no
         # number yet, addressed by its temporary tail.
         self.tail = scheme.temp_of(path)
--- a/luria/adr_index.py
+++ b/luria/adr_index.py
@@ -155,7 +155,9 @@ class Adr:
         self.path = path
         self.scheme = scheme
         self.prefix = scheme.prefix
-        self.number = scheme.number_of(path)
+        # Identity is the document's own `uid:` where it has one, the
+        # filename's number where it doesn't (#219).
+        self.number = scheme.uid_of(path)
         # A merge-allocated document awaiting concretization (ADR-049): no
         # number yet, addressed by its temporary tail.
         self.tail = scheme.temp_of(path)
--- a/luria/concretize.py
+++ b/luria/concretize.py
@@ -51,6 +51,7 @@ import sys
 from pathlib import Path
 
 from . import doc_refs
+from . import new as new_mod
 from .collect import _added_at
 from .config import current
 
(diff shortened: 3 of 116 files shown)


record/decisions.d/ADR-001.md
```diff
@@ -1,4 +1,5 @@
 ---
+uid: 1
```
dmarx: the 'u' in `uid` makes me want to assume that the ID is unique across the record, but it isn't: it's local to the scheme. so if the record has 3 schemes, each one well have its own `uid: 1`, so that uid actually resolves to three separate items if we consider it in isolation. I could see that potentially being a feature we might want in the future. what are some alternative ways we might name this field?

dmarx:
Good catch — renamed to `number:` and pushed (`0383669`). Alternatives I weighed, since you asked for them rather than a pick:

| | |
|---|---|
| **`number`** ✅ | What the code already calls this value — `number_of()`, `Adr.number`, the `{number}` URI template variable. Promises nothing about scope. |
| `seq` / `ordinal` | Accurate about the ordering property we chose it for over a UUID, but new vocabulary for a concept that already has a word. |
| `code: ADR-001` | Genuinely record-unique, so it answers your objection head-on rather than scoping around it, and it's what a citation looks like. But the prefix belongs to the *scheme*, not the document — storing it 400 times means renaming a scheme becomes a 400-file edit instead of one config line, and it invents a second drift class (prefix vs. directory). |
| `id` | Same over-promise as `uid`, and more overloaded. |

Two things turned up while checking that argue the same way:

- **`uid` was already taken.** A remote declares `uid` to mean an opaque foreign tail (`ADR-024`), so the same word would have named two unrelated things in one config file. I'd missed that.
- **Leaving `uid` unused keeps it free** for the record-global identifier you said the name made you expect — which seems worth reserving rather than spending.

Accessors follow the rename: `number_of()` is now the canonical reader (field first, filename fallback), and `number_in_name()` is the filename-only one that the lint and the repair need.

**The one real cost, since it's not free.** `number:` collides with English where `uid:` didn't — a line wrap in `LIT-158` of the consuming record already puts `number:` at column zero mid-sentence:

> …here mainly for the 10%
> `number: the record's optimizer practices trade convergence against…`

Reading only inside the frontmatter block is what makes that safe. The test for it now uses that real sentence instead of an invented one.

Re-verified after the rename: 981 tests pass, lint holds its baseline here, and the anthology migration re-run gives 436 documents with lint identical and exactly one view changed (`reference-status.md`, line numbers +1). I checked the committed views were current first, so the comparison means something — an earlier run of mine had disagreed and that turned out to be the confound.

---

record/decisions.d/ADR-001.md
dmarx: Renamed to `number:` — the word the code already used (`number_of()`, `Adr.number`, the `{number}` URI template variable). Two things turned up that argue the same way: `uid` was already taken (a remote's opaque foreign tail, `ADR-024`), and leaving it unused reserves it for the record-global identifier you said the name made you expect.

Full comparison of the alternatives, including `code: ADR-001` and why I rejected it, is in the PR conversation. Merged in `0383669`.

(Replying here late — GitHub was rate-limited when I first tried to answer on this thread.)

---

PR #221, opened 2026-09-08
A scheme can render an interpretable alias for its documents (#219)

dmarx:
Step 2 of #219, on top of #220.

```toml
[luria.schemes.LIT]
alias = "LIT-{first_author}-{published:.4}-{number}"
```

`LIT-Dao-2022-074` now resolves wherever `LIT-074` does — and **`luria link --fix` leaves it written.** That last part is the feature: a `formerly:` spelling gets rewritten away, and canonicalizing a derived one would erase the readable identifier on the first `--fix`, silently.

Both kinds live in one map and each entry carries its kind, because the fixer's instruction differs and nothing in a spelling's shape distinguishes them.

## The plan was wrong about where this goes

#219 said *"aliases join the resolver's map."* Checking that before writing against it was the useful hour: **`aliases.alias_map()` had no production caller** — two tests and nothing else. Resolution actually ran through `doc_refs.alias_number`, scanning every document's frontmatter per lookup.

Its docstring gave the sound reason for not caching:

> *"this path only runs for a temp code with no live document, which is rare"*

True for temporary codes, which nobody writes on purpose. **A spelling people choose to cite is not rare**, so the dead function turned out to be exactly the shape this needed, sitting one wire away. Wiring it up was the change; extending it was the easy part.

## Precision comes from the map, never the pattern

An alias tail has to be loose — words, digits, hyphens — and a loose pattern in a reference scanner matches prose. `SOTA-based` would become a citation.

The defence was already written down, in `legacy_spellings`: *"a temp-shaped string that resolves to no document is not a legacy spelling."* Same rule here, but **load-bearing rather than a nicety** — which makes the cache correctness, not just speed.

## Refused, and reported

- A template not starting with the scheme's prefix is **refused at load**: every scanner finds a code by its prefix first, so a spelling luria cannot see resolves for nobody.
- A collision is a **violation**, and deliberately not auto-suffixed — an automatic `-002` is order-dependent and would renumber when a third document arrives. `{number}` makes collisions impossible instead.

## Two measurements changed the design

- I probed the anthology for `authors:`, got zero, and **nearly reported the feature unusable on the record it was built for**. The field is `first_author:`, and 218 of 218 notes carry it. Reading one real file instead of trusting my own probe caught it.
- **The filter vocabulary I planned is mostly unnecessary.** `str.format`'s spec language truncates a *string*, so `{published:.4}` turns `2022-05-01` into `2022`. (It raises on an int, which is why an earlier probe made me think filters were required.) The motivating case needs none.

## Fired on the real record

218 literature notes rendered, **no collisions**, `LIT-Dao-2022-074` for FlashAttention — the pre-migration `MLR-2014-Kingma001-0001` shape recovered, which is the point of the exercise.

## Still to do, and named in the ADR

`luria repair` moving a superseded alias into `formerly:` when its source field changes. Until then an alias is safe to *resolve* but not yet safe to *write down durably* — correcting an author silently breaks citations written the old way.

## Diff

| | |
|---|---|
| `luria/aliases.py` | `Alias` with its kind, derived rendering, the map wired up |
| `luria/config.py` | `Scheme.alias`, validated eagerly |
| `luria/doc_refs.py` | resolution through the map; the fixer's two policies |
| `luria/lint.py` | `check_alias_collisions` |
| `tests/test_alias_inference.py` | 14 tests |
| `tests/test_migrations.py` | two assertions updated for the map's new shape |

**995 tests pass.** Lint exit 0, read whole — the two lines differing from `main` are this contribution's own opt-out directive and its own `Proposed` decision.

---


dmarx:
Pushed `63a5ba5` — **step 2 is now complete**. This was the piece the ADR listed as still to do: `luria repair` retiring a superseded alias into `formerly:`.

Without it an alias was safe to *resolve* but not to *cite durably* — correct an author and every citation written the old way breaks, silently.

## The old spelling comes from git, not a ledger

That was the design question. Storing each rendered alias in the document would give one copy per revision, which is precisely the hand-kept ledger `ADR-040` rejected. But the previous frontmatter is **already written down** — the record's history is the repository's — so the repair renders the alias from the last commit and compares.

Where git cannot answer (a new file, no repository) nothing is retired. **Absence of history is not a change of spelling**, and guessing there would write a `formerly:` naming something that never existed. There's a test for that case specifically, since a wrong guess would be invisible.

Once retired, the entry flips from AKA to FKA, so the same map that left it alone starts rewriting it — the two policies handing off exactly where they should.

## Also cleaned up

Replied on and resolved your `uid`/`number` thread on #220 — GitHub was rate-limited when I first tried.

**1000 tests pass**, lint unchanged from this branch's last green state.

---

Remaining on #219 is step 3 (one template vocabulary, two variable sources), and it shrank: `str.format`'s spec language already truncates strings, so the filter set I'd planned is largely unnecessary. #217's `derive` folds into it.

---

Issue #222, opened 2026-09-08
A test fixture is a quote, not an address: three suites now need `unlinted-file:`


dmarx:
Following through on a trigger I set myself on #220: *"Two files isn't a pattern; three would be."* It is three.

| file | why |
|---|---|
| `tests/test_migrations.py` | pre- and post-migration code spellings |
| `tests/test_number.py` | `ADR-007` (Superseded here), `ADR-tmpabcde` |
| `tests/test_alias_inference.py` | rendered alias spellings, `LIT-tmpabcde` |

Every one is the same situation: **a suite that tests how codes are read has to write codes down.** The reference scanner sees a citation; the author wrote a specimen.

Per the project's own rule — *a guard that keeps catching you is a bug report about the workflow; the fix is to remove what generates the mistake, not to keep thanking the net* — the fix is not a fourth directive.

## Options, roughly in order of how much I'd trust them

1. **Treat a test file as unlinted by default.** Simplest and matches the fact: not one of the three had a real citation to check. Costs the ability to catch a genuinely stale reference in a test's prose, which is a thing that could matter in a docstring explaining *why* a test exists.
2. **Scan tests, but only outside string literals.** A code inside `"..."` or `'''...'''` is data the test operates on; one in a comment or docstring is prose that can go stale. More precise, and it needs Python-aware scanning that the current text-based scanner does not do — the directive machinery deliberately works on raw text.
3. **Keep the directives and accept it.** Honest, and each one does carry a reason. But three identical acknowledgements in a row is what the rule above says to stop doing.

I'd lean 2 for correctness and 1 for cost, and I don't think this is mine to pick — option 1 quietly gives up a check, which is the sort of thing that should be decided rather than defaulted into.

Not urgent: the directives work and the record is clean. Filing it so the pattern is on the record rather than absorbed a fourth time.

dmarx:
4. come up with an annotation convention for acknowledging docstrings. can't we just do add an `inactive-ok: block` comment above the docstring?

dmarx:
Consolidated into #231 and closing here, so the decision is made once.

This is #132 arriving at a smaller scale — that issue measured the same hazard tree-wide in August and reached the same conclusion about the same mechanism. Both analyses are carried into #231, with the numbers re-measured on `main`.

Two updates since this was filed, both of which move option 2 from expensive to cheap:

[...]

PR #225, opened 2026-09-08
A docstring is one block, the way a fence already was (#222)

dmarx:
Your option 4 on #222 — *"can't we just add an `inactive-ok: block` comment above the docstring?"* Yes, and now you can:

```python
# inactive-ok-block: ADR-012 — the decision this replaced
def apply():
    '''First paragraph.

    A later paragraph citing ADR-012.'''
```

## It didn't work before — measured, not assumed

Block scope stopped at the docstring's first blank line. I tried the four placements on the real case:

| placement | before |
|---|---|
| no directive (the control) | warns |
| **`-block` above the `def`** | **reported STALE** |
| `-file` at module top | acknowledged |
| `inactive-ok:` inside the docstring | warns |

So only `-file` reached a citation in a later paragraph — far too blunt for one sentence of prose that happens to name a code. Three times today I hit this scope rule; the third time I **deleted the citation** rather than annotate it, which is the workflow telling you something.

## The fix was already written down, for a different case

The rules say fenced code counts as one block even when it contains blank lines — a syntactic unit is not split by the whitespace inside it. **A docstring is the same shape.** `blocks()` learns the file's language and treats docstring spans as atomic, counted from the `def`/`class` line so a directive above the definition governs what the definition says.

Four lines of real change. It's an existing principle applied to Python, not a new concept.

## Two false starts, both caught by the control

Worth recording because the lesson generalises:

- The first probe reported **every case quiet, including the control** — its replacement string never matched, so nothing was ever mutated.
- The second reported **every case acknowledged, including the control** — `luria lint` writes to stderr and I captured only stdout.

A probe whose control doesn't reproduce the condition is measuring nothing. Checking the control first is what caught both.

## What stays unchanged, deliberately

A directive written *inside* a docstring still does not fire — a docstring is not a comment. That rule exists because prose about the syntax used to invoke it, and a docstring example once silently annotated its own module. There's a test pinning it.

## Fired on the real case

The `ADR-087` citation I had deleted from `derive.value` is back, acknowledged by a `-block` directive above the definition.

## What this does *not* fix

The three suites carrying `unlinted-file:` are a different problem — fixture codes in **string literals and comments**, not docstring prose. #222's options 1–3 still stand for those, and that decision is still yours.

**1031 tests pass.** Lint identical to `main`'s baseline.

---

### DIFF
--- a/luria/directives.py
+++ b/luria/directives.py
@@ -32,7 +32,16 @@ defaults to remember:
 - **block** (`-block`) — the run of non-blank lines it sits in. A directive
   standing alone between blank lines has no content block of its own, so the
   block it means is the one it introduces: the next one. Fenced code counts as
-  one block even when it contains blank lines.
+  one block even when it contains blank lines, and so does a Python
+  docstring — one syntactic unit however many paragraphs it holds, counted
+  from its `def` or `class` line, so this reaches a citation in a docstring:
+
+      # inactive-ok-block: ADR-012 — the decision this function replaced
+      def apply(...):
+          '''First paragraph.
+
+          A later paragraph citing ADR-012.'''
+
 - **file** (`-file`) — the whole document.
 
 A blank line between a directive and what it governs therefore needs `-block`:
@@ -113,10 +122,50 @@ def _fence_line_spans(text: str) -> list[tuple[int, int]]:
     return spans
 
 
-def blocks(text: str) -> list[tuple[int, int]]:
-    """Blank-line-delimited runs of lines, 1-based inclusive. A fenced block is
-    atomic — a blank line inside a code sample doesn't end the paragraph."""
+def _docstring_line_spans(text: str) -> list[tuple[int, int]]:
+    """1-based (first, last) line numbers of each docstring in a Python source.
+
+    Module, class and function docstrings only — a string used as a value is
+    not one, and annotating it is not what anyone means. Unparseable source
+    yields nothing rather than raising: a directive scope is not the place to
+    discover a syntax error."""
+    import ast
+    try:
+        tree = ast.parse(text)
+    except (SyntaxError, ValueError):
+        return []
+    spans = []
+    for node in ast.walk(tree):
+        if not isinstance(node, (ast.Module, ast.ClassDef, ast.FunctionDef,
+                                 ast.AsyncFunctionDef)):
+            continue
+        body = getattr(node, "body", None)
+        if not body or not isinstance(body[0], ast.Expr):
+            continue
+        doc = body[0].value
+        if isinstance(doc, ast.Constant) and isinstance(doc.value, str):
+            # From the `def` line, so a directive written above the definition
+            # governs the docstring it introduces rather than stopping at the
+            # signature.
+            first = getattr(node, "lineno", doc.lineno)
+            spans.append((first, doc.end_lineno or doc.lineno))
+    return spans
+
+
+def blocks(text: str, path: Path | None = None) -> list[tuple[int, int]]:
+    """Blank-line-delimited runs of lines, 1-based inclusive.
+
+    A fenced block is atomic — a blank line inside a code sample doesn't end
+    the paragraph — and so is a **Python docstring**, for the same reason and
+    on the same principle (#222). A docstring is one syntactic unit however
+    many paragraphs it holds, so a `-block` directive written above a
+    definition governs the whole of what that definition says, not just its
+    first paragraph. Without it the only annotation that reaches a citation in
+    a docstring's third paragraph is `-file`, which is far too blunt for the
+    case: an ordinary sentence of prose that happens to name a code."""
     fenced = _fence_line_spans(text)
+    if path is not None and path.suffix.lower() == ".py":
+        fenced = fenced + _docstring_line_spans(text)
 
     def in_fence(line: int) -> bool:
         return any(a <= line <= b for a, b in fenced)
@@ -210,7 +259,7 @@ def _line_offsets(text: str) -> list[int]:
 
 def find(path: Path, text: str, names: set[str] | None = None) -> list[Directive]:
     """Every directive in `path`, with the lines each one governs resolved."""
-    spans = blocks(text)
+    spans = blocks(text, path)
     found: list[Directive] = []
     for line_no, offset, body in comment_fragments(path, text):
         m = DIRECTIVE_RE.match(body.strip())
--- a/docs/devlog/2026-09.md
+++ b/docs/devlog/2026-09.md
@@ -2,7 +2,7 @@
 
 # Development log — September 2026
 
-42 entries. [All books](README.md).
+43 entries. [All books](README.md).
 
 ## Contents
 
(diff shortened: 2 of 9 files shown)


luria/directives.py
```diff
@@ -113,10 +122,50 @@ def _fence_line_spans(text: str) -> list[tuple[int, int]]:
     return spans
 
 
-def blocks(text: str) -> list[tuple[int, int]]:
-    """Blank-line-delimited runs of lines, 1-based inclusive. A fenced block is
-    atomic — a blank line inside a code sample doesn't end the paragraph."""
+def _docstring_line_spans(text: str) -> list[tuple[int, int]]:
```
dmarx, line 125: rather than a bespoke python-specific solution here, could we just use tree-sitter for language-generic support?

dmarx: Measured it rather than guessing, and the answer splits: **not for this, but you're pointing at a real bug next door.**

## For the docstring case, tree-sitter buys almost nothing

I ran six documentation styles through `blocks()`:

| style | one unit? |
|---|---|
| Python docstring | ✅ (this PR) |
| JSDoc / Javadoc `/** * */` | ✅ already |
| Rust `///` run | ✅ already |
| Go `//` run | ✅ already |
| Shell `#` run | ✅ already |
| C `/* */` with a **truly blank** line | ❌ still |

Four of the six already worked, by an accident of syntax worth noting: their separator lines carry a comment marker (` *`, `///`, `//`, `#`), so they were never blank and the run never split. **The problem only exists where documentation contains a genuinely empty line** — a Python string literal, or a C-style block comment.

So it isn't a Python-specific fix so much as a fix for the one common language that documents in string literals. The remaining gap is one shape, and it needs a delimiter table, not a parser.

And tree-sitter wouldn't remove the per-language knowledge anyway — you'd still need a map from language to *"which node is the documentation unit"*, expressed as node-type names instead of delimiters. Same table, different spelling, plus ~3 MB of platform-specific compiled wheels (`tree_sitter` 665 KB + `tree_sitter_language_pack` 2.4 MB) against two pure-Python dependencies today, in a package installed in every consumer's CI.

## But the crude scanner has a real correctness bug, and that's where it would earn its keep

Chasing your question I found this, which is worse than the thing you asked about:

```yaml
- run: echo "see # inactive-ok: ADR-012 — not a directive"
```

**That fires a directive.** So does `// inactive-ok:` inside a Go raw string, and `#` inside a shell string. The fallback scanner is line-based and reads string literals as comments.

This is live in luria's own tree — `code_globs` includes `.github/workflows/*.yml`. And it's precisely the failure the module's own rule was written against: *"comments in the scanner explaining these rules registered as directives."* A directive can currently be invoked from inside a string.

Python and Markdown are exact already (stdlib `tokenize`, HTML comments). It's everything else that's approximate.

## What I'd propose

1. **Land this PR as is** — it fixes the case that actually bit, with no dependency.
2. **File the string-literal false positive separately** — it's independent of docstrings and more serious.
3. **Tree-sitter as an *optional extra*** for exactly the languages on the crude scan, degrading to today's behaviour when absent. That gets exactness where it's missing without putting compiled wheels in the required install path, and leaves the already-exact paths alone.

Happy to do 2 and 3 now if you'd rather see the shape before landing this. Your call on whether the optional-extra split is worth it or whether you'd sooner just take the dependency outright.

---

dmarx: **You're right and I was wrong.** I argued that from priors instead of measuring it. Installed tree-sitter and ran your construction — comment node, then next named sibling — across seven languages:

| language | comment node type | next named sibling | governs |
|---|---|---|---|
| python | `comment` | `function_definition` | lines 1–7 |
| javascript | `comment` | `function_declaration` | 1–7 |
| go | `comment` | `function_declaration` | 1–4 |
| rust | `line_comment` | `function_item` | 1–4 |
| yaml | `comment` | `block_node` | 2–6 |
| bash | `comment` | `function_definition` | 1–4 |
| c | `comment` | `function_definition` | 1–4 |

`"comment" in node.type` covers every one — including Rust's `line_comment`, which is the case I expected to need a table for. **That's one substring rule, not a per-language map.** My "same table, different spelling" claim was simply false.

And the docstring question doesn't survive it either. It isn't *"which node is the documentation unit"* — the Python row governs lines 1–7, which **includes the docstring**, because the docstring is inside the function the directive introduces. The problem I spent this PR solving dissolves rather than being solved: you don't need to know what a docstring is, only what the next sibling is. The JavaScript row covers the `/* */`-with-a-blank-line case for the same reason, which was the one gap I'd said was left.

## And a correction to #226, which I got wrong

My headline example there was **not a bug**:

```yaml
- run: echo "see # inactive-ok: ADR-012 — fake"
```

That `#` genuinely *is* a YAML comment — the value is a plain scalar, so the quotes are ordinary characters and the comment starts at ` #`. PyYAML parses the value as `'echo "see'`. The crude scanner was right, tree-sitter agrees with it, and my "live in this repository" claim rested on that example.

The bug is real, but it needs the *properly quoted* form:

```yaml
- run: "echo see # inactive-ok: ADR-012 — inside a quoted scalar"
```

| | luria today | tree-sitter |
|---|---|---|
| quoted YAML scalar | **fires** ❌ | no comment node ✅ |
| `#` in a bash string | **fires** ❌ | none ✅ |
| `//` in a Go raw string | **fires** ❌ | none ✅ |
| a real bash comment | fires ✅ | found ✅ |

So the false positive stands on three of four, with a corrected demonstration. I'll fix #226's example.

## What I'd do now

Build it your way: a generic `-block` resolved as *comment node + next named sibling*, which replaces both the docstring special case in this PR and the crude scanner's string-literal guessing.

The one call I'd still leave to you is packaging — `tree_sitter` (665 KB) plus `tree_sitter_language_pack` (2.4 MB) are platform-specific compiled wheels against two pure-Python dependencies today, in something that installs into every consumer's CI. I'd default to an **optional extra** that degrades to today's behaviour when absent, but if you'd rather have it required and exact everywhere that's one line, and I'll do whichever you say.

Starting on it now on that assumption.

---

dmarx:
Built: #227, stacked on this branch.

Your rule, as specified — a comment node plus the construct it introduces — with one correction the measurement forced. Taking the comment's *next named sibling* is right in Python, Go, Rust, JS, bash and C, and wrong in YAML, where a top-level comment's next sibling is the whole document: `-block` would quietly become `-file` in every workflow file. What #227 takes instead is the smallest named node with children that starts where the block's content starts and holds the whole block. Same answer everywhere your rule was right, and the `jobs:` entry rather than the document where it wasn't.

Stated as a union with the blank-line block, so a grammar can only widen what a directive governs — a record that lints today cannot start failing because the extra got installed.

**On this PR's fate:** I'd land it first rather than fold it in. The `ast` docstring rule stays as Python's fallback when the extra is not installed, which is what most installs will be, and #227's tests assert the two paths agree on the docstring case. If you'd rather require the grammar, the special case goes and that is a one-line change — but I would want the pin below settled upstream first.

**The pin.** `tree-sitter` 0.26.0 corrupts the heap when trees are walked. It cost most of the work, because the crash moved with every unrelated edit and kept surfacing in `ast.parse` in a worker thread; I chased and wrote comments for three wrong causes (the traversal, threads, the markdown grammar) before building the thing that settled it — ~25 lines using nothing from luria, one grammar, a cursor walk over this repo's own `.py` files. Segfaults on 0.26.0, clean on 0.25.2. The extra pins `<0.26` and `LURIA_TREE_SITTER=0` exists for the next such release.

I also corrected my #226 example on the issue: `run: echo "see # …"` genuinely *is* a YAML comment, so luria was right to fire there. The real case is the quoted form, and #227 silences it along with the bash and Go ones.

---

PR #239, opened 2026-09-11
lint: reject HTML comments and duplicate keys in YAML frontmatter

### LINKED ISSUE #164: Lint: reject HTML comments inside YAML frontmatter (PyYAML accepts them as keys; Quartz does not)
## What happened

In dmarx/anthology-of-the-sota#15, an acknowledgement directive was written into a practice's frontmatter as an HTML comment:

```yaml
published: '2024-09-01'
<!-- inactive-ok: LIT-141 — the superseded paper, named as the predecessor in the chain -->
source: LIT-141
summary: >-
<!-- inactive-ok: LIT-141 — the superseded paper, named as the predecessor in the chain -->
  Zhu et al. (2024), ...
```

PyYAML parses each `<!-- inactive-ok: LIT-141 — ... -->` line as a mapping key (`<!-- inactive-ok`) with a string value, and silently keeps the last of the two duplicates. So `luria lint` reported the document clean, `luria index` rendered it, and the record's own frontmatter check passed. Quartz's YAML parser rejected the file with `duplicated mapping key`, and the downstream Pages build went red.

The correct spelling is the `#` comment that 0.8.1 reads in frontmatter (#163); the mistake was a script inserting the markdown form without checking where it was.

## Proposal

Two checks, both mechanically decidable, so they belong in the failing lint rather than a report:

1. **A frontmatter line that opens with `<!--` is a finding.** There is no legitimate reason for an HTML comment inside YAML, and the failure it causes is downstream and silent here.
2. **Duplicate mapping keys in frontmatter are a finding.** PyYAML's default loader takes the last value; a strict loader (or a `construct_mapping` override that raises on repeats) catches the case generally, of which (1) is the most likely instance.

Optionally, `luria repair` could rewrite `<!-- inactive-ok: ... -->` inside frontmatter to `# inactive-ok: ...`, since the intent is unambiguous.

## Also noticed

`luria lint` on `main` reports one pre-existing legacy spelling that `luria link --fix` does not upgrade: `record/decisions.d/ADR-075.md:90  ADR-tmpxmnac → ADR-071`. Not touched in #163.

HarshRajSinghania:
## Summary

`luria lint` now rejects two frontmatter shapes that PyYAML accepts and Quartz rejects:

1. A frontmatter line that opens with `<!--` (treated as a mapping key by PyYAML).
2. Duplicate top-level mapping keys (PyYAML keeps the last value).

A YAML `#` comment is still allowed. List items and nested keys are not treated as top-level keys.

## Motivation

Fixes #164. A practice in anthology-of-the-sota had `<!-- inactive-ok: … -->` written into YAML frontmatter. `luria lint` reported the file clean; Quartz failed the Pages build with `duplicated mapping key`.

## Implementation

- New helper `luria/frontmatter_shape.py` inspects the raw `---` block before `yaml.safe_load`.
- `check_frontmatter` runs that helper on every scheme document (including temp documents).
- Changelog fragment under `record/changelog.d/`.

Repair of `<!-- inactive-ok -->` to `# inactive-ok` is not included; the issue listed that as optional.

## Testing

```
pip install -e ".[dev]"
python -m pytest tests/test_frontmatter_shape.py tests/test_lint.py tests/test_examples.py tests/test_statuses.py -q
```

Result: 138 passed (after the list-item false-positive on `- version:` was fixed).

Also ran the three new tests in isolation: all passed.

### DIFF
--- a/luria/frontmatter_shape.py
+++ b/luria/frontmatter_shape.py
@@ -1,0 +1,51 @@
+"""Frontmatter shape checks that PyYAML's default loader will not raise.
+
+PyYAML treats a line that opens with `<!--` as a mapping key and keeps the
+last of two identical keys. Quartz (and other strict YAML parsers) reject
+both, so `luria lint` used to report a document clean while the downstream
+Pages build went red (#164).
+"""
+
+from __future__ import annotations
+
+
+def frontmatter_raw(text: str) -> str | None:
+    """The YAML block between the opening and closing `---` fences, or None
+    when the document has no well-formed frontmatter. Kept as raw text so a
+    check can see what PyYAML silently rewrites."""
+    if not text.startswith("---\n"):
+        return None
+    end = text.find("\n---\n", 3)
+    if end == -1:
+        return None
+    return text[4:end + 1]
+
+
+def check(errors: list[str], rel: str, text: str) -> None:
+    """Reject HTML comments and duplicate mapping keys in YAML frontmatter."""
+    block = frontmatter_raw(text)
+    if block is None:
+        return
+    seen: dict[str, int] = {}
+    for line in block.splitlines():
+        stripped = line.lstrip()
+        if stripped.startswith("<!--"):
+            errors.append(
+                f"{rel}: HTML comment in YAML frontmatter — use a `#` "
+                f"comment (PyYAML treats `<!--` as a mapping key)")
+        # Top-level mapping keys only: unindented, not a YAML `#` comment
+        # or a list item, and carrying a colon. Nested keys are indented.
+        if (not line or line[0].isspace() or stripped.startswith("#")
+                or stripped.startswith("-")):
+            continue
+        if ":" not in line:
+            continue
+        key = line.split(":", 1)[0].rstrip()
+        if not key:
+            continue
+        seen[key] = seen.get(key, 0) + 1
+    for key, count in seen.items():
+        if count > 1:
+            errors.append(
+                f"{rel}: duplicate frontmatter key {key!r} "
+                f"(PyYAML keeps the last value; strict parsers reject it)")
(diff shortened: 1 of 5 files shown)


luria/frontmatter_shape.py
```diff
@@ -0,0 +1,51 @@
...
+
+def check(errors: list[str], rel: str, text: str) -> None:
+    """Reject HTML comments and duplicate mapping keys in YAML frontmatter."""
+    block = frontmatter_raw(text)
+    if block is None:
+        return
+    seen: dict[str, int] = {}
+    for line in block.splitlines():
+        stripped = line.lstrip()
+        if stripped.startswith("<!--"):
```
dmarx: we should probably prefer `line.startswith` rather than `stripped.startswith` here. column 0 is what makes PyYAML read it as a key, and an indented `<!--` inside a `summary: >-` is legal content. After a quick scan I wasn't able to locate any live examples like this, so this is prophylactic, but its likely the first instance will be the entry explaining this bug.

dmarx (changes requested):
small change requested, also please submit a devlog fragment to accompany the PR

luria/frontmatter_shape.py
HarshRajSinghania: Updated: the HTML-comment check now uses `line.startswith("<!--")` so only column-0 comments are rejected. An indented `<!--` inside a folded scalar is left alone.

Also added a devlog fragment at `record/devlog.d/2026/09/11/054100.md`, plus a test covering the indented case. `pytest tests/test_frontmatter_shape.py` — 4 passed.

dmarx:
Thanks for the contribution! The upstream issue was two complaints, one of which you addressed here. Didn't want to delay integrating your improvements so accepted your PR as is and closed the old issue, and created an independent self-contained issue for the other half if you wanna pick that up. If not, thanks anyway for contributing!

https://github.com/dmarx/luria/issues/240

PR #245, opened 2026-09-12
A cited document gets a page, and citations point at it

dmarx:
> Split out of [#244](https://github.com/dmarx/luria/pull/244) and rebased onto `main`, so it can land while the strata-g viewer integration is still being workshopped. None of the graph/viewer work is here.

## The bug

Every citation of a design principle resolved to `docs/design-principles.md#dp-N`. The anchors that names are `<a name="dp-3"></a>` — raw HTML, correct as markdown, and preserved only by a publisher that chooses to preserve it. Quartz does not: its remark→hast pipeline drops the element and slugifies each heading's own text instead. So a link written against an anchor the source *genuinely contains* lands on a page that has no such id.

Measured on a real v4.5.2 build of this record: **330 links across 81 pages, and not one of them resolved.**

They work in the repository and fail on the site, which is the worst place for a difference to live. Every check this project runs reads markdown, and `docs/design-principles.md` resolves perfectly as a path.

## Two halves, and the first one alone makes things worse

This PR started as the configurable half only — a `cite` key, default left at today's behaviour. It passed everything, rewrote nothing, and fixed nothing, which @dmarx spotted immediately by running `pages.yml` off the branch. Dogfooding it is what found the rest:

> setting `cite = "page"` with the old `publishable()` took links redirected to the repository from **10 to 195**

A page target names `record/principles.d/DP-003.md`, and that file was not published — so the site did the only thing it could and sent every one of those readers out to GitHub. That is a *downgrade* from the bug: a dead anchor at least lands you on the principles document with the content on it, at the wrong scroll position.

`luria lint` cannot see this. It checks that a relative target exists **on disk**, and it does. Staging is the only thing that reads a link as something a reader would try to follow.

### 1. A document-rendered scheme's sources are published as pages

`record/principles.d/DP-004.md` is served at `/record/principles.d/DP-004`, alias `/DP-004`, exactly like a decision. The assembled view is still published — both, like a decision and its index.

`publishable()` excluded them by a *derived* rule: `link_base(path) != path.parent` asks "are this file's links spelled for somewhere else?" The derivation is good. What was wrong is that a design principle answers yes for a completely different reason than a changelog fragment does — a fragment has no code, no title of its own, nothing to give a page **to**; a principle is numbered, titled, statused, versioned and cited by code. Every property a decision has except an address.

So the rule now says what it meant:

| | Published? |
|---|---|
| A **fragment** — no code, no title of its own, no identity apart from the view | No, unchanged |
| A **document** that happens to be rendered as a section of one | **Yes, now** |

The exception is to the *consequence*, not the rule: those links really are spelled for `docs/`, so `_rebase` re-points each relative target as the page is written. The repository spelling is untouched, because the fixer and the lint both depend on it — `luria lint` would fail the moment it stopped.

### 2. `[luria.schemes.X] cite` chooses where a citation points

```toml
[luria.schemes.DP]
render = "document"
output = "docs/design-principles.md"
cite   = "page"          # or "view"
```

`"page"` resolves to the cited document's own file, `"view"` to `output#anchor`. **Unset means what the scheme already does** — `"view"` for a document scheme, `"page"` for an index scheme, which has no second address — so the key is inert for a project that never sets it. **This record sets `"page"`.**

Two refusals, because they are different mistakes. An unknown word is a typo. An *explicit* `cite = "view"` on an index scheme is a request that cannot be honoured — there is no assembled document to anchor into — and resolving to the page anyway would answer a question the project did not ask.

### `luria repair` moves the links a record already wrote

Without this the key would be half a feature: flipping it governs every citation written from then on and **nothing already on disk**, because those are plain markdown links and the linkifier spells *bare* references. It rewrote **98 files** here.

`retarget_view_citations` is deliberately narrow:

- the link **text** is left alone — it is the author's sentence, not a field;
- a link with **no fragment** is left alone — pointing at the whole assembled document is a real thing to do;
- an anchor **naming no document** is left alone — rewriting it would swap a dead fragment for a dead *file*, which is worse and hides it from the lint.

## Verification

**On real Quartz v4.5.2 builds of this record, either side of the change:**

| | before | after |
|---|---|---|
| broken anchor links | **330** across 81 pages | **0** |
| links to a principle's page | 85 | **404**, all resolving |
| staged pages | 281 | **307** |
| links redirected to the repository | 10 | **12** |

That last row is the one worth reading hardest. It is how you know the new pages **absorb** the new targets rather than leaking them off-site: the same config change without half 1 made it 195.

`/DP-004` and `/record/principles.d/DP-004` both answer. `python -m pytest tests -q` → **1101 passed, 33 skipped**. `luria lint` clean.

**19 cases cover the key** (`tests/test_scheme_cite.py`), pinning every direction through **both** resolvers — `wikilink_target` for `[[DP-2]]` and `resolve` for a bare prose `DP-2` — since they are separate code paths that must agree and the output does not say which one wrote a link. Including an instrument check: the fixture's assembled view really does carry `<a name="dp-2">` and `dp_anchors()` really does find it, so a test that prefers the page is **choosing** rather than falling back.

**Two tests changed their claim rather than their threshold**, which is the part to review:

- `test_publishable_is_exactly_the_files_whose_links_resolve_in_place` asserted `link_base(path) == path.parent` over the whole corpus — the same claim as the rule, while no exception existed. It now exempts a document source and says why.
- `test_a_nested_record_is_staged_by_its_own_config` proved its point by the child's document-scheme source being **absent** from the parent's staging, which this legitimately makes present. It proves the same point by the presence of the code **alias** instead: only a config that knows the child's scheme can mint one.

Neither was relaxed to make the build green.

## Picked up along the way

Two links were spelled `design-principles.md#6-fire-before-trusting` — a heading slug no rule here has ever emitted, so `retarget_view_citations` correctly refused to guess at them. Rewritten as `[[DP-6]]` so `luria repair` spells the target, rather than hand-writing a third path in a file where hand-written paths are precisely the hazard.

## Changelog

- [x] `record/changelog.d/claude-cite-target.md`

## Devlog

- [x] `record/devlog.d/2026/09/12/215952.md` — the 10 → 195 experiment, because *"a link target is only better than another if the thing it names is published"* is the part worth not rediscovering
- [x] Adds an ADR — the generated views come from CI's `docs-generate` job

## Reading the diff

130 files, but only **9 are code**: `luria/site.py`, `luria/config.py`, `luria/doc_refs.py`, `luria/link_refs.py`, `luria.toml` and four test files. Of the rest, **98 are record sources whose citations `luria repair` rewrote** and 20 are regenerated views.

## The other side of the split

[#244](https://github.com/dmarx/luria/pull/244) keeps the strata-g viewer integration: the typed lineage graph, the configured site graph, the record map and its ego slicing. Once this lands it rebases onto `main` and sheds everything above.

https://claude.ai/code/session_01YSRQzpm4kAWQSaKWD6V9x2

---


dmarx:
I ran the Pages workflow (pages.yml) off this branch, and the DP links still resolve to `output#anchor`. maybe we still need to dogfood this change in the repo's own luria.toml?

dmarx:
That's the PR working as intended, and you've put your finger on the thing it doesn't do.

`cite` unset resolves to what the scheme already does, which for a `render = "document"` scheme is `view` — the anchor. So a build off this branch is byte-identical to a build off `main`. That's deliberate (it's what makes the PR verifiable as isolated: 1097 tests pass with none modified, `luria repair` writes 0 files), but it does mean **this branch fixes nothing you can see.**

## Dogfooding it here makes the site worse, measured

I tried exactly what you suggested — `cite = "page"` in this repo's `luria.toml`, `luria repair`, then stage:

| | links redirected to the repository |
|---|---|
| `cite` unset — what you built | **10** |
| `cite = "page"` on this branch | **195** |

`luria repair` rewrote 98 files, and 185 principle citations left the site. A `page` target points at `record/principles.d/DP-003.md`, and on `main` **that file is not published** — a document scheme renders its sources into one view, and the view is what gets a page. So staging does the only thing it can and sends the reader to the GitHub source.

That's a downgrade from the bug you're looking at, not a fix. The dead anchor at least lands you on the principles document with the content on it, at the wrong scroll position; the `page` target takes you off the site to raw markdown.

## So the key needs the pages, and the pages are in #244

`cite = "page"` is only the better target once a document scheme's sources *are* pages. That change — `document_source()`, the one-line narrowing of `publishable()`, `_rebase()` to re-spell a source's links on the way out — is commit `7aff8a1` on #244's branch, and it turns out to be self-contained: the only parts of it that depend on the viewer work are a `lineage_index` hunk and a `test_site_graph.py` hunk, and neither exists on `main`.

Two ways to go, and it's your call:

1. **Lift `7aff8a1` onto this branch.** #245 becomes self-contained and actually fixes what you saw: principles get pages, `cite = "page"` set here, the ~330 anchor citations repointed, verified on a real Quartz build (330 broken → 0 when I did it on the stack). Bigger PR than you asked for, but none of it is the strata-g viewer integration you're still workshopping — that stays entirely in #242/#243/#244.
2. **Ship #245 as plumbing.** Inert by design, unblocks the flip, and the pages change lands separately.

I'd take (1) — it's the version where the PR's own motivation is demonstrable rather than described, and it leaves the graph work untouched. Say which and I'll do it.


---

dmarx:
3. instead of lifting a commit, just do the change fresh in the config and run `luria repair` or whatever as needed, and commit the changes. like I said: dogfood please.

dmarx:
Done — `6498882`. Written fresh against `main`'s `site.py`, not lifted.

You were right to push, and dogfooding found something I'd have missed: **the key alone makes the site worse**, and only staging can see it. Setting `cite = "page"` with the old `publishable()` took links redirected to the repository from **10 to 195**, because a page target names `record/principles.d/DP-003.md` and that file wasn't published. `luria lint` can't catch it — it checks that a relative target exists *on disk*, and it does.

So the change is two halves now:

**1. A document-rendered scheme's sources are published as pages.** The exclusion was a derived rule — "are this file's links spelled for somewhere else?" — and the derivation is fine. What was wrong is that a principle answers yes for a different reason than a changelog fragment does: a fragment has no code, no title of its own, nothing to give a page *to*; a principle is numbered, titled, statused, versioned, cited by code. The rule now says what it meant, and the exception is to the **consequence** — those links really are spelled for `docs/`, so `_rebase` re-points them as the page is written and the repository spelling is untouched (the fixer and the lint both depend on it).

**2. `cite = "page"` set in this repo's `luria.toml`.** `luria repair` rewrote 98 files.

## On a real Quartz v4.5.2 build, either side

| | before | after |
|---|---|---|
| broken anchor links | **330** across 81 pages | **0** |
| links to a principle's page | 85 | **404**, all resolving |
| staged pages | 281 | **307** |
| links redirected to the repository | 10 | **12** |

That last row is the one I'd look at hardest — it's how you know the new pages *absorb* the new targets rather than leaking them off-site. 195 was the failure; 12 is the fix.

`/DP-004` and `/record/principles.d/DP-004` both answer. 1101 tests pass, `luria lint` clean.

## Two tests changed their claim, not their threshold

- `test_publishable_is_exactly_the_files_whose_links_resolve_in_place` asserted `link_base(path) == path.parent` over the whole corpus — the same claim as the rule while no exception existed. It now exempts a document source and says why.
- The nested-record test proved its point by the child's source being **absent** from the parent's staging, which this legitimately makes present. It proves the same point by the presence of the code **alias** instead: only a config that knows the child's scheme can mint one.

Neither was relaxed to make the build green.

## One thing the fixer wouldn't do

Two links were spelled `design-principles.md#6-fire-before-trusting` — a heading slug no rule here has ever emitted, so `retarget_view_citations` correctly refused to guess. I wrote `[[DP-6]]` and let `luria repair` spell the target rather than hand-writing a third path in a file where hand-written paths are the hazard.

The ADR and changelog fragment are rewritten for what actually shipped; the devlog entry records the 10 → 195 experiment, since "a link target is only better if the thing it names is published" is the part worth not rediscovering.


---

PR #258, opened 2026-09-14
An anchor is an id

dmarx:
Reported as "the devlog links appear to be broken on the rendered site". They are, and so is every other fragment link this record publishes.

## One attribute

The generator anchored a journal entry with `<a name="20260914042025"></a>` and an assembled document with `<a name="dp-3"></a>`. A `name` is reached only by a **real navigation** — the HTML spec's fallback to `a[name]`. Quartz is a single-page app; its router scrolls with

```js
document.getElementById(decodeURIComponent(url.hash.substring(1)))
```

(v4.5.2, `quartz/components/scripts/spa.inline.ts`, both call sites), which finds an `id` and nothing else.

**89 of the 100 fragment links in this repository resolved by `name` and no other way** — every devlog entry link on every index (55 in September alone), and every citation of a design principle. All of them landed at the top of the page they named.

Verified against the live site rather than reasoned about: the published `2026-09` page contains all 55 `<a name="…">` elements and zero matching `id`s. Nothing is dropping the element.

The shape of the failure is the thing to take from it. It was **correct everywhere a contributor would check and wrong only where a reader would read**: every check luria runs reads markdown, the path resolves, the anchor is genuinely in the file, the link is genuinely spelled right. Nothing short of loading the built site could see it.

## What changed

**Both emitters write `id`.** Strictly a widening — `id` is reachable everywhere `name` was, plus by `getElementById`. `doc_refs` still *reads* either spelling, because a project whose principles are one hand-written file may anchor them by `name` and that file resolves fine where it is read; refusing it would turn a publishing defect into a parsing one.

**A check, with a `--fix`.** `luria lint` reports a fragment link whose target answers to it by `<a name=>` alone; `luria link --fix` rewrites the anchor. The generator is not the only thing that writes one, and the next hand-written anchor would put this back silently.

Two things the check had to get right, and got wrong first:

- **It reads views.** Every other reference rule reads sources, correctly — a view is rewritten by the next build, so a finding about one is a finding nobody can act on where it is reported. But the motivating case is a journal *index* linking into a journal *book*, and both are generated. Pointed at the pre-fix book, the first version found **nothing**: a check shaped so it cannot see the defect it exists for. It finds 55 now. `doc_files(views=True)` is the exception and the anchor check is its only caller.
- **It resolves from `link_base`, not the linking file's own directory.** A `render = "document"` scheme's prose is written to resolve from the page it assembles into, so `../../docs/values.md` in a source is correct there and nonsense from the source's folder. Naive resolution reported five correct links in the constitution example as broken targets.

**The repair is the target's, and its owner decides the remedy.** The link is spelled correctly; the thing it names cannot be found. So `--fix` edits the document holding the anchor — and never a view, because the next `luria index` would erase it. A finding about a view says `luria index` instead.

## ADR-094 is corrected in place and versioned

Its measurement held — *330 links across 81 pages, none of which resolved*. Its stated cause did not: it recorded that Quartz *"drops the element and slugifies each heading's own text instead"*, and rejected **"emit anchors that survive the publisher"** on that basis, as a fix needing re-finding for each publisher.

Quartz drops nothing. The rejected alternative was the right answer and one attribute wide. Corrected per this record's own rule (ADR-001: the choice stood, the reason was wrong) rather than superseded — the decision itself still holds, and `cite` keeps its other reason, which was always the real one: a section read out of its context reads oddly.

An explanation that fits the evidence is not the same as an explanation that is true. "Quartz drops raw HTML" fits "none of these links resolve" exactly. Checking it cost one `curl` and a `grep`.

## Verification

- **1191 passed, 33 skipped, 0 failed.** Eight new in `tests/test_anchors.py`, including the view→view case a source-only scan cannot see.
- `luria lint` clean, 0 unacknowledged references.
- Re-audited the corpus after regenerating: 89 `ok: explicit id`, 6 `ok: heading slug`, 0 name-only.
- The check run against the pre-fix book from `main` reports all 55, each naming `luria index` as the remedy.
- Generated views reverted before committing (ADR-018); `luria index` repairs an existing record's views in one pass, and nothing in a source has to change.

## Two things left deliberately

**`cite: page` stays as configured.** DP citations currently point at source files, which *work*. Flipping DP back to `cite: view` now that anchors resolve is a choice about where a citation should land, not a bug fix.

**The anthology has the same breakage** — its published `docs/design-principles` carries 9 `name`-anchors and no matching ids. It is still on `luria.toml` pinned at `0.15.0`, so it needs `luria upgrade yaml` and an unpin before it picks this up.

https://claude.ai/code/session_01BuxSz1G5aEoaQYfKSToenz

---
_Generated by [Claude Code](https://claude.ai/code/session_01BuxSz1G5aEoaQYfKSToenz)_


dmarx:
## Green — `f98727c`. The CI failure was the useful part of this PR.

`docs-check` reported **55 violations from my own new check**, and it was right to.

### What I did wrong

The check read the committed views off disk. Locally it was green — because I had run `luria index` before `luria lint`, so the views on disk were the fixed ones. CI checks out the branch and lints *without* generating, and a branch carries the **default branch's** copies of every view, because this record deliberately does not commit views on a branch (ADR-068).

So the check reported 55 violations against files this PR is **forbidden to fix**, and would have done so on every future PR that touched an anchor.

Worse: I had written the wart into the ADR and argued it was fine.

> the check reads views, so it is the one reference rule that can report a finding a contributor cannot fix where it is reported. That is the right trade here

It was not a trade. It was a bug I rationalised because I could not see how else to reach a view→view link.

### The answer was already in the codebase, one function away

`check_view_dirs`:

> computed against what the generator would write, in memory, so the check reads sources and writes nothing … **a branch carries the default branch's copies and has nothing to be stale against.**

So `anchors.documents()` hands the check every source plus `adr_index.outputs()` — the views **as this source tree renders them**. The question stops being *"what is committed"* and becomes *"what will this record produce"*, which is the only version with the same answer on a branch and on `main`. `doc_files(views=True)` had no other caller and is gone.

**Verified both directions this time**, which the first round skipped:

| condition | findings |
|---|---|
| fixed generator + `main`'s stale views on disk (CI's exact state) | **0** |
| source fix stashed so `main`'s generator runs | **55** |

### Two smaller repairs fell out of it

- **A stub is the authored part of a generated page**, so an anchor written in one was found twice — in the stub and in the view it renders into — and the view copy's suggested remedy was wrong, because a person *can* fix a stub. One finding per unreachable fragment now, reported where somebody can act on it.
- **`--fix` keyed the repair on the link's target**, which for a stub anchor is the view. It repairs every authored document carrying a reported fragment instead. Loose, and safe to be: only anchors some link actually named are touched, and `name` → `id` is an improvement wherever it lands.

Also fixed a stale `inactive-ok-file:` that CI flagged in the same run — the directive vouched for a code the file never cited. Now it does.

### Verification

**`test` ✅ · `docs-check` ✅** — 1191 passing locally, `luria lint` clean with 0 unacknowledged references.

ADR-tmp1wx7r's Decision and Consequences are rewritten against what the check actually is, and the devlog entry carries the full account — including that this was the *third* shape of the check: sources-only couldn't see the defect, disk-views couldn't be trusted on a branch, and the render is the only reading that is right on both counts.

---

PR #274, opened 2026-09-16
A vocabulary and a tag group can carry an alert, printed with their own violation

### LINKED ISSUE #273: Let a rule carry its own note, so a closed vocabulary can say "add a value" instead of "pick from this list"
A closed vocabulary is the right guard and the wrong message. Its violation text reads as *this value is not allowed*, when the intended reading is often *this value has not been declared yet, and declaring it is the move*. There is currently no way for a record to say which it means.

## Concretely

`anthology-of-the-sota` just closed `tags` on three schemes so that every tag is one somebody chose and blurbed — this matters there because the chain invariant now reads `tags`, so an undeclared word typed once can silently bind a relation. Today a contributor reaching for a word the vocabulary lacks gets:

```
record/practices.d/SOTA-085.md: `tags: made-up-tag` is not in the `topics` vocabulary
(vocabulary 'topics') — the values are training-optimization, systems-optimization,
model-stability, distributed-optimization, data-pipeline, attention-techniques,
model-architecture, inference-optimization, adaptation-and-tuning,
representation-and-encoding, analysis-and-evaluation, generative-modeling,
vision-and-graphics, tiny-models
```

Correct, and it teaches the wrong lesson. The path of least resistance it sets up is *pick the nearest of these fourteen* — which is exactly the failure the record has been fixing all week: eight positional-encoding papers filed under `model-architecture` and `attention-techniques` because `representation-and-encoding` did not exist yet, and nobody went back after it was added.

That record's owner put it well:

> I don't mind setting the vocabulary closed as an enforcement mechanism to move new tags into the official structured vocabulary, but I'm concerned that the enforcement mechanism will mainly discourage the addition of new tags.

The guard wants to be a turnstile, not a wall, and nothing in the config can express the difference.

## Proposal

Let a rule carry a note that the violation prints:

```yaml
vocabularies:
  topics:
    closed_note: >-
      Closed so every tag is one somebody chose, not because the list is
      finished. If this document needs a word the vocabulary lacks, add it
      here with a label and a blurb and a decision saying why — that is the
      intended move, not a workaround.
```

rendered as:

```
record/practices.d/SOTA-085.md: `tags: made-up-tag` is not in the `topics` vocabulary … 
  ↳ Closed so every tag is one somebody chose, not because the list is finished. If this
    document needs a word the vocabulary lacks, add it here with a label and a blurb and a
    decision saying why — that is the intended move, not a workaround.
```

## Shape

The general version is more useful than the specific one: **any rule that can fail should be able to carry a `note:`**, appended to its own violation — `closed`, `required`, `required_when`, a tag group's `require`, a reference's `required`. Each is a place where a record has a reason that the generic message cannot carry, and where the reason changes what a contributor does next.

`Field` already threads `because` for provenance and `_cite(field.because)` renders it, so there is a precedent for carrying config-side explanation into a message; a `note` would ride alongside it. `contract._vocabulary_violations` is the single call site for this one.

Two details worth deciding rather than guessing:

- **Where it attaches.** On the vocabulary (`vocabularies.topics.closed_note`) reads best for this case, since the remedy is "edit this table". On the field (`schemes.SOTA.fields.tags.note`) generalizes better, since the same vocabulary may be closed on one scheme and open on another — which is true in this very record, where `NOTE.tags` declares no vocabulary at all.
- **Whether it is one note or one per rule.** `note:` on the field covers everything it can fail; `closed_note:`/`required_note:` are precise but multiply.

I'd lean field-level with per-rule keys, but the call is yours.

## Why not just do it in prose

The record's `CLAUDE.md` now carries the invitation, which helps a reader who is reading `CLAUDE.md`. It does not help the one who typed a tag, ran the lint, and is looking at a message telling them fourteen values are available. The note wants to be where the failure is.

## Related

- #272 — the cross-scheme chain limitation, from the same pass in the same record. The invariant change that made closing worth doing is the reason that one matters too.

dmarx:
Closes #273.

A closed set's finding says what is allowed. Whether the list is **finished** or merely **short** is a fact only the record knows, and it had nowhere to say it — so the message reads as *pick one of these*, which is how a vocabulary stops growing.

## The config

```yaml
fields:
  tags:
    vocabulary: topics
    closed: true
    alert: >-
      Closed so every tag is one somebody chose, not because the list is
      finished — add a value here rather than reaching for the nearest
      wrong one.
```

```
record/practices.d/SOTA-085.md: `tags: kv-cache-paging` is not in the `topics`
vocabulary (vocabulary 'topics') — the values are training-optimization, …
    ↳ Closed so every tag is one somebody chose, not because the list is
      finished — add a value here rather than reaching for the nearest wrong one.
```

`TagGroup` takes the same key and prints it on its `require` findings and on `excluded_by`, which is that rule stated from the other side.

## The name, after review

It was `note` while this was being built. Your review asked for a word that says *message emitted alongside a linter finding*, offering `warning` or `alert`.

**`alert`**, and the reason is a collision: luria already grades findings as warnings and violations, and `lint.fail_on` promotes one to the other. A key called `warning:` on a vocabulary would read as a severity dial. What this holds is the sentence printed *when the rule fires*, and the finding it rides is a violation — so naming it for a severity it doesn't set would be wrong twice over. Say the word if you'd rather have `warning` anyway; it's a one-line change plus the tests.

Separately, your point about `blurb` on the vocabulary itself is filed as #279 — the gap is real and wider than vocabularies (`TagGroup`, `PlainField`, `Scheme` and `FieldGroup` have none either, while relations and vocabulary *values* do). The two keys are complementary rather than competing: `blurb` describes the thing at rest, `alert` is what it says when you trip over it. A vocabulary would reasonably carry both.

## Not the per-rule keys the issue proposed

#273 suggested `closed_note:` / `required_note:` on the field. That's rejected, and the reason is the useful part — `ADR-tmppr2mz` records it.

A **field** fails several ways: absent, wrong shape, outside the closed set. One alert on a field would print *"add a value to the vocabulary"* on a missing-field error, where it's simply wrong — so a field-level alert needs a per-rule key just to be correct, and then there's one key per rule per field.

A **vocabulary** has one thing to explain: its values, and whether the set is closed. A **group** has one `require`. Put the alert on those and each has exactly one failure to explain, so the per-rule spelling disappears. The general form: *an alert belongs with the rule, and a rule belongs to something smaller than a field.*

## Tests — 6 new, 1285 passing

Two of the six assert the **absence** case directly: a vocabulary or group without an alert produces byte-identical messages to before. One asserts an alert doesn't leak from one group onto another group's finding — that one caught a wrong assertion I'd written, where I'd asserted `strength`'s alert on a `failure` violation; the behaviour was right and my test was wrong.

## Fired on the real case, per the working agreement

Against `anthology-of-the-sota` — the record that asked for it — not only a fixture. The devlog entry carries the output. Two things noticed alongside, neither this change's business:

- **The closed-set finding is emitted twice.** Identically on the released 0.22.0 with my branch uninstalled, so it predates this and is not a regression here. The scheme's contract holds no duplicate `tags` field, so it comes from somewhere in the scan. Worth a look; not looked at.
- **An older luria ignores the new key** rather than erroring, so a record can adopt it before the release that reads it and get the old message meanwhile. Checked against 0.22.0 rather than assumed.

## Record entries

Changelog fragment, devlog entry, and a decision document — the alternative rejected was real and was the one the issue proposed. The decision now also carries a short section on the name. `luria lint` exits 0; generated views and the README badge region are deliberately not committed, per the views-on-main-only policy.

## What it doesn't reach

`FieldGroup.require`, `required_when`, and a reference's `required` still produce messages a project cannot annotate. They weren't the case asked for, and each would want the same treatment on the same argument — the carrier, not the field.


dmarx:
This is actually highlighting a separate issue, which is that we should add a blurb field to objects like vocabularies so the user can explain its role/purpose/meaning wrt the record. right now the user can independently describe fields that invoke the vocabulary and individual objects within it, but not the vocabulary as a whole. at least I think that's the case, right? I think we might have blurbs for relations maybe, but if that's not a thing either we should definitely add that. 

with blurbs available on these other objects, I'd reframe the feature we're proposing here from a "note" field to a "warning" or "alert" field to emphasize that this is a message that gets emitted alongside linter findings.

dmarx:
Both right. Audited against the dataclasses rather than from memory:

| object | `label` | `blurb` |
|---|---|---|
| `Reference` (a relation) | ✅ | ✅ — #254, so your "maybe" is a yes |
| a vocabulary **value** | ✅ | ✅ |
| `Journal` | — | ✅ |
| `Chain` | `title` | — |
| **`Vocabulary`** | — | — |
| **`TagGroup`** | — | — |
| **`PlainField`** | — | — |
| **`Scheme`** | — | — |
| **`FieldGroup`** | — | — |

So the hole is where you said, and wider: you can say what `extends` means and what `training-optimization` means, but not what the `topics` vocabulary *is* — and the same is true of tag groups, plain fields, schemes and field groups. Filed as #279, with the note that it's #254's own argument one level up: a vocabulary's purpose is currently a comment nothing can render, quote or scaffold from. In `anthology-of-the-sota` that's ~35 lines of prose explaining the thirteen topics, reachable by no view.

**Renamed to `alert`, pushed.** I took your first option's intent but not the word: luria already grades findings as warnings and violations, and `lint.fail_on` promotes one to the other, so `warning:` on a vocabulary reads as a severity dial. This holds the sentence printed *when the rule fires*, and the finding it rides is a violation — so naming it for a severity it doesn't set would be wrong twice. Say the word if you'd rather have `warning` anyway and I'll flip it; it's a one-line change plus the tests.

```yaml
topics:
  closed: true
  alert: >-
    Closed so every tag is one somebody chose, not because the list is
    finished — add a value here rather than reaching for the nearest wrong one.
```

The two keys are complementary rather than competing, which is worth stating before #279 lands: `blurb` describes the thing at rest, `alert` is what it says when you trip over it. A vocabulary would reasonably carry both.

1285 tests pass, `luria lint` clean. The decision document now carries a short section on the name.

---

PR #286, opened 2026-09-16
Rewrite the example configs too (code.globs)

dmarx:
Cleaning up after myself: #285 left a live temporary code on `main`.

`luria concretize` rewrites the files `code.globs` names, and this repository's own list doesn't name `examples/**/*.yaml`. So ADR-111 landed as `ADR-tmp92495` in `examples/knowledge-base/luria.yaml` — the `.py` files were rewritten correctly, the example config wasn't.

**Nothing caught it, and nothing could.** A temporary code goes on resolving through the successor's `formerly:` entry, which is exactly what that field is for, so `luria lint` is clean and the unresolved-codes report is unchanged. The only symptom is a shipped example reading as though it were authored halfway through a branch.

Two changes:

- `examples/**/*.yaml` added to `code.globs` — brings 7 example configs into the rewrite set
- the stale `ADR-tmp92495` → `ADR-111`

### Verified

Before/after on the collected file set: `examples/knowledge-base/luria.yaml` goes `False` → `True`.

And end to end, because the file set being right doesn't prove the rewrite happens — a scratch ADR cited from the example config, then `luria concretize`:

```
ADR-tmp5mzvb → ADR-112
```

with the example rewritten from `(ADR-111, ADR-tmp5mzvb)` to `(ADR-111, ADR-112)`. Scratch reset afterwards; ADR-112 is not in this branch.

Full suite 1335 passed / 33 skipped. No views committed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FeX27AYgHU7geZaYkwshBS)_


dmarx:
Checked whether `anthology-of-the-sota` has the same gap. It doesn't — and it doesn't because it already hit this and fixed it the same way. Its `code.globs` names `luria.yaml` explicitly, with this in the config:

> It was scanned by nothing, so those claims went stale silently: `ADR-035` was cited 8× here after it had become ADR-035, and **four more temp codes besides**. A reference the lint cannot see is a reference nobody maintains (ADR-039, *"The configuration is part of the record, and the lint reads it"*).

Same failure, same cause, one file class over: a decision-dense file outside `code.globs` accumulates stale codes in silence, because every one of them keeps resolving.

That's an argument for going further than this PR does. I've added `examples/**/*.yaml` because that's where the live temp code actually landed, but the general shape is *any file that cites decisions and isn't in the list* — and for this repo that plausibly includes `README.md` and the `record/*/README.stub` files. I haven't audited those and didn't want to widen the PR on a guess; say the word and I'll check what citations they carry before proposing anything.

---

